PR #16, opened 2021-04-02
model extension: unique keys

cmungall:
This introduces compound keys 

need to add tests

this example should illustrate concept:

```
classes:
  isotope:
    slots:
      - id
      - atomic number
      - neutron number
      - symbol
   compound_keys:
     proton_neutron_key:
       unique_for: isotope
       primary: false
       compound_key_slots:
         - atomic number
         - neutron number
    symbol uniqueness:
      unique_for: isotope
       primary: false
      compound_key_slots:
         - symbol

slots:
   id:
    identifier: true
```

See https://github.com/linkml/linkml/issues/42

### DIFF
--- a/model/schema/meta.yaml
+++ b/model/schema/meta.yaml
@@ -505,6 +505,23 @@ slots:
     range: boolean
     description: indicator that this is the root class in tree structures
 
+  compound_keys:
+    domain: class_definition
+    range: compound_key_definition
+    multivalued: true
+    inlined: true
+    description: Set of compound keys for this slot
+    mappings:
+      - owl:hasKey
+  
+  compound_key_slots:
+    domain: compound_key
+    multivalued: true
+    required: true
+    range: slot_definition
+    description: list of slot names that form a key
+    comments:
+      - We allow compound keys with a single compound_key slot, but at least one must be specified
 
   # -----------------------------------
   # Slot definition slots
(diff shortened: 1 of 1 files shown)


hsolbrig:
We should draft a prototype of what compound keys looks like in each of the emitted idioms:
* markdown - how to document
* python - what does a compound key look like in generated python and what contract(s) does it enforce
* owl -
* rdf -
* json schema -
* others?

cmungall:
good Qs, rough sketch of answers:

markdown

should pretty much follow the yaml. Each compound key on one line. E.g.

```

### Compound Keys

1. proton_neutron_key: an isotope is uniquely identified by the combination of (atomic number, neutron number)
2. symbol_key: an isotype is uniquely identified by the symbol field

```

python

this may not be easy to validate at the python level

rdf/owl

trivial mapping to hasKey

jsonschema

no equivalent?

hsolbrig (commented):
I wonder whether we should use this as an opportunity to create a new mixin that has the set of slots and mixins that you've enumerated below.  I think we ended up with a similar set in the enumerations and I'd prefer to minimize repetition and fragmentation.

If you agree, will be happy to make the proposal

cmungall:
regarding nul//none values in the key:

We could have slot to indicate whether nulls are to be treated as unknown vs a special None value. If the former, then it's not valid to translate the unique_key to SQL unique constraint (since Nulls represent unknowns)

PR #26, opened 2021-06-18
Adding recommended field

cmungall:
fixes https://github.com/linkml/linkml/issues/177


hsolbrig (approved):
Looks good.  We may want to  disallow "recommended: true / required: true"

PR #29, opened 2021-06-18
Clarifying docs on id_prefixes and adding explicit documentation on elipses ... syntax

cmungall:


hsolbrig (commented):
`...` is not a valid ncname, so we'd have to change the range to include it.   While I don't have a strong opinion,  I weakly favor explicit semantics (e.g. `"open": true` vs. "if '...' in prefix list...

cmungall:
OK - we said `...` on a call in April, see https://github.com/linkml/linkml/issues/194#issuecomment-830374311

but happy to go with other scheme

hsolbrig (changes requested):
I think we should remove the elipses comment line until we actually know what and how we are going to implement it.  As is this pull commits us to some immediate tasks

cmungall:
ok, removed the comment on elipses until we decide on open vs closed prefixes

hsolbrig (approved):

PR #32, opened 2021-06-26
Adding min and max cardinality of multivalued slots

cmungall:


hsolbrig (commented):
While I don't object to this construct, it is going to involve a number of changes to the generators and loaders, as well as reconciliation w/ the multiple/required elements.

Out of curiosity, is there a compelling use case for this?  Observations have been that:
1) min and max, while ubiquitous,  end up adding complications to wherever they show up.  They have to be reconciled w/ required, multiple, and one has to come up with some sort of compound data type to support unlimited max, and all of this involves a lot of edge cases and detail work.
2) besides the hypothetical test cases, they are almost _never_ used, outside of the already doable [0..1], [1..1], [0..\*] and [1..\*].

One should also note that, were we to cover all the bases, we would want to be able to distinguish between the absence of a list and a list with no members -- meaning that we would want to separate the semantics of `required` from the `min` element.  Not suggesting we should do this, just pointing it out if we decide to go ahead.

PR #33, opened 2021-07-14
Refactor model schema

hsolbrig:
This PR proposes that we move the model/schema into the linkml_model directory, the reason being that we want to have the models themselves as part of the pypi install.  Currently we do this by copying ./model/schema to ./linkml_model/model/schema, but this has resulted in a lot of confusion, with folks (including me) editing the wrong schema more often than not...


hsolbrig:
Oops - we need to get the linkml_files.py as a special case in the linkml_model distro -- do NOT pull yet

PR #45, opened 2021-09-20
add title to all definitions

cmungall:


gaurav:
This is probably a dumb question, but why “title” rather than “name” or “label”? I never know which one of those I should use, and would appreciate a deeper understanding of how you made that choice here.

Issue #59, opened 2021-11-19
what the heck is 11179 with respect to permissible value meanings?


turbomam:

turbomam:
Add documentation about [this](https://en.wikipedia.org/wiki/ISO/IEC_11179) to ???

gaurav:
There might be some information about this (at least in the CCDH context) is at https://cancerdhc.github.io/terminology/ -- I believe @hsolbrig and @jiaola wrote this document.

PR #60, opened 2021-11-19
#59 move mention of ISO 11179 from description to see_also

turbomam:
permissible_value.meaning.Description was
> the value meaning (in the 11179 sense) of a permissible value

I moved the mention of ISO 11179 into a see_also of https://en.wikipedia.org/wiki/ISO/IEC_11179

### DIFF
--- a/linkml_model/model/schema/meta.yaml
+++ b/linkml_model/model/schema/meta.yaml
@@ -324,9 +324,11 @@ slots:
     domain: permissible_value
     range: uriorcurie
     description: >-
-      the value meaning (in the 11179 sense) of a permissible value
+      the value meaning of a permissible value
     notes:
       - we may want to change the range of this (and other) elements in the model to an entitydescription type construct
+    see_also:
+      - https://en.wikipedia.org/wiki/ISO/IEC_11179
 
   # -----------------------------------
   # Schema definition slots


turbomam:
for issue #59

gaurav:
I'll defer to @hsolbrig on this definition, but note that the "value meaning" in the ISO 11179 sense is apparently "The meaning or semantic content of a value" as per https://wiki.nci.nih.gov/display/caDSR/ISO+11179+Term+Definitions. So potentially you could change the definition to read: `The meaning or semantic content of a value (i.e. it's "value meaning" according to the ISO 11179 standard).` I tried to find a clearer definition in [ISO 11179-5](https://www.iso.org/standard/50340.html), but I was more confused than enlightened :)

hsolbrig:
I like Guarav's proposal.  I fret about going from a reasonably well defined term -- "Value Meaning" as defined by 11179 to "value meaning" -- whatever you think it means.  No problem with adding a citation (it should be 11179-3, not 11179-5), but I think we loose if we fail to say where we got a particular term and make it our own.

There is quite a bit of detail in section 11.3.2.3.2 in ISO_IEC_11179-3_2013.pdf (http://standards.iso.org/ittf/PubliclyAvailableStandards/c050340_ISO_IEC_11179-3_2013.zip)

cmungall (approved):

PR #64, opened 2021-12-01
Add alias object

sierra-moxon:


linkml_model/model/schema/aliases.yaml
```diff
@@ -0,0 +1,84 @@
...
+    required: true
+    description: >-
+      Can be a URIOrCurie that represents the source of the alias, or it can be a simple string that
+      describes the source of the alias.
+
+  alias_scope:
+    range: alias_scope_enum
+
+  alias:
+    domain: slot_definition
```
cmungall, line 45: not just for slots

cmungall: ah I see, it's already this way in the metamodel

alias is a very special purpose slot in the metamodel... unfortunately we have overloaded "alias" a lot...

linkml_model/model/schema/aliases.yaml
```diff
@@ -0,0 +1,84 @@
...
+    range: string
+    slot_uri: skos:prefLabel
+    description: >-
+      the name used for a slot in the context of its owning class.  If present, this is used instead of the actual
+      slot name.
+
+  aliases:
+    range: alias_definition
```
cmungall, line 53: unfortunately this would be a breaking change, we will need to have a different slot name here, like structured_aliases

yes, this means there are potentially duplicative ways of assigning aliases, but I think we have to live with that until we move to linkml2.0

linkml_model/model/schema/aliases.yaml
sierra-moxon: this is a slightly odd existing slot in the model already (I just moved it from meta.yaml to alias.yaml).  It's for setting a 'preferred label' of a slot.  alias_definition is what I named the 'alias class' in order to not break this model.

linkml_model/model/schema/aliases.yaml
```diff
@@ -0,0 +1,84 @@
...
+    multivalued: true
+    inlined: true
+    description: >-
+      A list of alias_definition objects.
+
+classes:
+  alias_definition:
+    is_a: element
```
cmungall, line 61: alternative: don't model as an element, instead inline as a list

linkml_model/model/schema/aliases.yaml
```diff
@@ -0,0 +1,84 @@
...
+      name:
+        exact_mappings: skos:altLabel
+        description: >-
+          inherits name from element class, but in the alias_definition class, this name represents
+          a skos:altLabel
+
+
+enums:
+  alias_scope_enum:
+    permissible_values:
```
cmungall, line 78: and exact!

linkml_model/model/schema/aliases.yaml
```diff
@@ -0,0 +1,84 @@
...
+      narrow:
+        meaning: skos:narrowMatch
+      broad:
+        meaning: skos:broadMatch
+      close:
```
cmungall, line 83: and related too (not the same as close)

linkml_model/model/schema/aliases.yaml
```diff
@@ -0,0 +1,84 @@
...
+  - OIO
+  - IAO
+
+imports:
+  - linkml:types
+  - linkml:meta
+
+slots:
+  alias_source:
+    range: string
```
cmungall, line 35: uriorcurie?

linkml_model/model/schema/aliases.yaml
cmungall: we might want to keep the common_metadata mixin though,

linkml_model/model/schema/aliases.yaml
```diff
@@ -0,0 +1,84 @@
...
+  - rdf
+  - rdfs
+  - xsd
+  - skos
+  - OIO
+  - IAO
+
+imports:
+  - linkml:types
+  - linkml:meta
```
cmungall, line 31: so modularization is quite tricky here, I think meta will want to import this, rather than vice versa.. I don't know if there are reciprocal dependencies

but to keep it simple, maybe the alias model can just live in meta

cmungall:
I see 32 files changed - intentional?

PR #66, opened 2021-12-10
add alias suggestion

sierra-moxon:


linkml_model/model/schema/meta.yaml
```diff
@@ -1333,6 +1360,15 @@ classes:
     close_mappings:
       - qb:HierarchicalCodeList
 
+  alias_definition:
```
cmungall, line 1363: so our naming conventions are a bit odd. Only slot_definition and class_definition inherit from definition... yet schema_definition inherits from element, despite it's name...

How about we reserved Xdefinition for things with identifiers, we can simply call this structured_alias?

linkml_model/model/schema/meta.yaml
```diff
@@ -1333,6 +1360,15 @@ classes:
...
+    description: >-
+      object that contains meta data about a synonym or alias including where it came from (source)
+      and its scope (narrow, broad, etc.)
+    slots:
+      - alias_name
+      - alias_source
+      - alias_scope
```
cmungall, line 1370: scope: This naming convention is very obo-format esque and not really used elsewhere. I would favor alias_type (perhaps alias_property/predicate/category)

linkml_model/model/schema/meta.yaml
```diff
@@ -1333,6 +1360,15 @@ classes:
     close_mappings:
       - qb:HierarchicalCodeList
 
+  alias_definition:
+    description: >-
+      object that contains meta data about a synonym or alias including where it came from (source)
+      and its scope (narrow, broad, etc.)
+    slots:
+      - alias_name
+      - alias_source
```
cmungall, line 1369: Just added `source` #62 -- could be reused?

linkml_model/model/schema/meta.yaml
```diff
@@ -109,9 +109,37 @@ slots:
...
+
+  alias_name:
+    range: string
+    required: true
+    exact_mappings: skos:altLabel
+    description: >-
+      inherits name from element class, but in the alias_definition class, this name represents
+      a skos:altLabel
+
+  structured_aliases:
```
cmungall, line 133: We need to also add this as a class_slot - I think under `element` is the right place, as that already has `aliases`

linkml_model/model/schema/meta.yaml
sierra-moxon: done.

linkml_model/model/schema/meta.yaml
sierra-moxon: I removed the 'alias_' convention and this turned into 'type' (and changed 'alias_scope_enum' to 'type_enum').

linkml_model/model/schema/meta.yaml
sierra-moxon: moved to use 'source' instead of 'alias_source.'  I'm a little worried that we won't have uriorcurie for every alias source (eg.: I have sources in the biolink-model that are 'fake' curies to get around this requirement in the mapping objects like 'GAMMA:activator' where GAMMA is a made-up prefix), but, it could be that we change the alias source in this case to an actual source that might be broader than the made-up source.

linkml_model/model/schema/meta.yaml
sierra-moxon: done

linkml_model/model/schema/meta.yaml
```diff
@@ -1346,6 +1359,22 @@ classes:
     close_mappings:
       - qb:HierarchicalCodeList
 
+  structured_alias:
+    description: >-
+      object that contains meta data about a synonym or alias including where it came from (source)
+      and its scope (narrow, broad, etc.)
+    slots:
+      - name
```
cmungall, line 1367: ah, unfortunately this is problematic. The metamodel follows a tradition in logic/linked data in using "name" to designate an identifier - and indeed `name` is marked as being an identifier, which will cause problems here, e.g. it will be inlined as a dict with the "names" being keys.

A perfectly acceptable alternate would be `alias` but that is actually slot-specific and used to designate the preferred label for the slot (i.e. unmangled)

Naming things is hard, I might just go with `alias_value` or `label`

linkml_model/model/schema/meta.yaml
```diff
@@ -110,9 +110,22 @@ slots:
     in_subset:
       - owl
 
+  type:
```
cmungall, line 113: I suggest `alias_type` OR keeping type generic and putting the specifics in slot_usage... but in this case I think alias_type and alias_type_enum is the way to go

cmungall (changes requested):
also can you add an example to the examples/ folder - see the Makefile for how to test

wdduncan:
@sierra-moxon @cmungall and everyone else ...  
Here is the link to the [alias enums in linkml](https://docs.google.com/document/d/1RnxFHLABmZxYfTFgwR1moxrX3Qu7831a2NC2quxz3BQ/edit) google doc.

linkml_model/graphql/meta.graphql
```diff
@@ -625,6 +625,13 @@ interface SlotExpression
     allOf: [AnonymousSlotExpression]
   }
 
+type StructuredAlias
+  {
+    name: String!
+    source: Uriorcurie
```
wdduncan, line 631: Will there be cases in which the source is identified simply with a string instead of a uri or curi?

linkml_model/model/schema/meta.yaml
```diff
@@ -110,9 +110,22 @@ slots:
...
+    range: type_enum
+    description: >-
+      The relatedness of an alias to the element it is aliasing.  type can be: narrow, broad, exact,
```
wdduncan, line 116: Instead of just `narrow` and `broad`, do you want `is narrower than` and `is broader than` ? It may make it clearer.

linkml_model/model/schema/meta.yaml
wdduncan: I'm not sure what the best approach is here. Could you also have subproperties of `alias_type` such as:
- `has narrow alias`
- `has broad alias`
- `has exact alias`
- `has close alias`
- etc.

From querying standpoint the difference would be:
- Find all alias names where the alias_type = exact
vs.
- Find all values of has exact alias

PR #75, opened 2022-01-05
Revert "make extension_tag a key fixes https://github.com/linkml/linkml/issues/548"

cmungall:
Reverts linkml/linkml-model#73


cmungall:
Need to think through effects of this

Declaring this as a key has the result of changing meta.py to also import AnnotationTag:

https://github.com/linkml/linkml-model/blame/main/linkml_model/meta.py#L24

https://github.com/linkml/linkml-model/blame/main/linkml_model/meta.py#L212-L213

This is problematic if different versions are combined

PR #79, opened 2022-02-01
adding structured patterns, adding ranking

cmungall:
- Added slot rank, for ordering of elements. Proposed model for structured patterns, addresses https://github.com/linkml/linkml/issues/176
- illustrates https://github.com/linkml/linkml/issues/176
- expanding examples

### DIFF
--- a/examples/pattern-example.yaml
+++ b/examples/pattern-example.yaml
@@ -1,0 +1,86 @@
+id: https://w3id.org/linkml/examples/pattern
+title: Pattern Example
+name: pattern-example
+description: This demonstrates the use of patterns
+license: https://creativecommons.org/publicdomain/zero/1.0/
+
+prefixes:
+  linkml: https://w3id.org/linkml/
+  ex: https://w3id.org/linkml/examples/patterns/
+  sh: https://w3id.org/shacl/
+
+default_prefix: ex
+default_range: string
+
+default_curi_maps:
+    - semweb_context
+
+emit_prefixes:
+  - linkml
+  - rdf
+  - rdfs
+  - xsd
+  - owl
+
+imports:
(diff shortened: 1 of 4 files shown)


examples/pattern-example.yaml
```diff
@@ -0,0 +1,86 @@
...
+
+#==================================
+# Slots                           #
+#==================================
+
+slots:
+  id:
+    identifier: true
+    range: string
+    pattern: "^P\\d{7}"    ## simple pattern. Note {}s are NOT interpolated
```
ddooley, line 56: I see now that the simple "pattern" allows flexibility in start or end of string matching.  Meanwhile as you have it, the structured_pattern, involving string components, can't have re's that have start or end markers.

examples/pattern-example.yaml
```diff
@@ -0,0 +1,86 @@
...
+  name:
+    range: string
+  email:
+    range: EmailString
+  height:
+    range: string
+    structured_pattern:
+      pattern: "{float} {unit.length}"
```
ddooley: Technical question, can a structured_pattern pattern have a mixture of {setting} and re expressions, like  "{float}\s*{unit.length}" ?

ddooley: Side note, it seems like for javascript based programming, at runtime load of a linkml spec, each given "pattern" can be converted into a compiled regular expression by expanding it out and then re.compile() ?  I've done this for very efficient runtime validation of data.

cmungall: Q1: I had envisioned yes, and that most implementations would interpolate directly to a regex.

JS: yes, that's the idea. And stay tuned re JS and linkml...

PR #94, opened 2022-04-29
aliases in common_metadata

turbomam:
see https://github.com/linkml/linkml/issues/732


sierra-moxon (approved):

cmungall (changes requested):
can you do the same for 

      - structured_aliases
      - mappings
      - exact mappings
      - close mappings
      - related mappings
      - narrow mappings
      - broad mappings
      - rank

also remove from under Element, no longer needed there are element already inherits from CM

cmungall (approved):

PR #95, opened 2022-05-03
proposal: add 'if_absent' as a list_value_specification_constant to add default values to a slot, fixes #633

sierra-moxon:


cmungall (changes requested):
A good suggestion.... however we already have something similar:

```yaml
  ifabsent:
    domain: slot_definition
    range: string
    inherited: true
    description:  |-
      function that provides a default value for the slot.  Possible values for this slot are defined in
      linkml_runtime.utils.ifabsent_functions.default_library:
        * [Tt]rue -- boolean True
        * [Ff]alse -- boolean False
        * int(value) -- integer value
        * str(value) -- string value
        * default_range -- schema default range
        * bnode -- blank node identifier
        * slot_uri -- URI for the slot
        * class_curie -- CURIE for the containing class
        * class_uri -- URI for the containing class
    close_mappings:
      - sh:defaultValue
```

and this is heavily used in the metamodel...

cmungall:
see also https://github.com/linkml/linkml/issues?q=ifabsent

PR #99, opened 2022-05-19
Delete README.md

cmungall:
Thanks to @julesjacobsen for reporting

The README is misleading as this is in fact the files that should be edited


julesjacobsen:
\m/

PR #120, opened 2022-08-11
adding publisher slot

cmungall:


matentzn:
I am beginning to think that a dataset-level FAIR LinkML core model shared around all data models would have been great. Something based a lot on void and dcat3

cmungall:
> I am beginning to think that a dataset-level FAIR LinkML core model shared around all data models would have been great. Something based a lot on void and dcat3

https://github.com/cmungall/data-manifesto

PR #121, opened 2022-08-24
Devops fixes for repo

sujaypatil96:
* poetrification of repo
* Makefile targets for `gen-project` and `deploy` modified
* GitHub Action that automatically builds and publishes documentation based on modifications to the schema


Makefile
```diff
@@ -42,14 +42,17 @@ install:
 
 all: gen-project gendoc
 %.yaml: gen-project
-deploy: all mkd-gh-deploy
+deploy: gendoc mkd-gh-deploy
```
sujaypatil96, line 45: Why rerun artifact generation step once again while doing `deploy`?

We just need the HTML docs in the right directory for deploying, so we can make gendoc a dependency on the deploy target

Makefile
```diff
@@ -42,14 +42,17 @@ install:
...
 
 # generates all project files
+# and updates the artifacts in linkml-model
 gen-project: $(PYMODEL)
 	$(RUN) gen-project -d $(DEST) $(SOURCE_SCHEMA_PATH) && mv $(DEST)/*.py $(PYMODEL)
+	cp -r $(DEST)/* $(PYMODEL)
+	rm -r $(PYMODEL)/docs
```
sujaypatil96, line 52: Copy over all the latest artifacts from `target/docs`. Otherwise there is no point in generating all the artifacts in the first place?

sujaypatil96 (commented):
Left some notes.

Makefile
```diff
@@ -42,14 +42,17 @@ install:
...
 
 test:
-	$(RUN) gen-project -d tmp $(SOURCE_SCHEMA_PATH) 
+	$(RUN) python -m unittest discover
```
cmungall, line 55: does this test changes to the schema?

cmungall (approved):

PR #129, opened 2022-11-15
Remove owner property.

hsolbrig:
Fixes issue https://github.com/linkml/linkml/issues/39


cmungall (commented):
This would be a breaking change - should we not first remove all uses of this slot in the main linkml repo first?

PR #130, opened 2022-11-17
Fix link in navigation to SpecificationSubset.

dalito:
Should the rebuilt docs also be committed or are they build in a CI job? I could not yet build them on windows.


sierra-moxon:
should only be pushed to gh-pages branch via actions, no the main repo.

PR #137, opened 2022-12-12
Update ifabsent documentation.

hsolbrig:
Documentation said "str", while actual pattern was "string"


cmungall (approved):

hsolbrig:
Actually, it would probably behoove us to add some testing code.  Hold off on the actual pull

hsolbrig:
The documentation now fits the model, and there is a test case to see how it works.  Note that several of the functions don't currently work...

cmungall (approved):

PR #139, opened 2022-12-21
update the tests copy of the model files with the test copy

sierra-moxon:
it looks like this is still a necessary step when we update the model, is that right?


cmungall (approved):
Looks good! We should probably review the need for a duplicate copy in tests at some point...

PR #146, opened 2023-01-24
Fix uri slot value for date type

vincentvialard:
bring uri value in sync with notes, just like time a few lines above


cmungall (changes requested):
Marking as request changes until we can do a thorough review, this will have implications as these types are used for mapping

PR #149, opened 2023-02-03
Adding a curie type

cmungall:
approved on linkml develper call 2023-02-03

@hsolbrig - this is the initial commit on types.yaml: https://github.com/linkml/linkml-model/blob/f9d1fa4ce407189e3a27eafbad53fbf52a2487e6/linkml_model/model/schema/types.yaml

### DIFF
--- a/linkml_model/model/schema/types.yaml
+++ b/linkml_model/model/schema/types.yaml
@@ -121,14 +121,20 @@ types:
     comments:
       - in RDF serializations this MUST be expanded to a URI
       - in non-RDF serializations MAY be serialized as the compact representation
+    conforms_to: https://www.w3.org/TR/curie/
 
   uri:
     uri: xsd:anyURI
     base: URI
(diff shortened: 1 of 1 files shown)


hsolbrig (approved):

hsolbrig:
Before we integrate this, I think we need to define the behavior of this type.

Questions: 
1) is ":x" a valid Curie?  how about "x" w/o the colon.
2) is "ex:x" a valid Curie if "ex" is not in a prefix map?

cmungall:
Added a `conforms_to` to the W3C spec.

Note we already have an implementation of this, we just made the corresponding entry in types (we have used uriorcurie for cases were we need curies)

https://github.com/linkml/linkml-runtime/blob/a8b4dd14abab2eca48cd10a99b287077fcc2c55c/linkml_runtime/utils/metamodelcore.py#L148-L188

PR #167, opened 2023-08-18
basic repo maintenance

sierra-moxon:
- I added a Makefile target to generate Python classes for all our meta-model YAML files (instead of only running `gen-project` which does `meta.yaml` with `-mergeimports=False`).  
- I also noticed the codespell config was ignoring all the yaml files with the regex it used to stop spell-checking on generated `.HTML` files.  I removed that regex and limited it by file extension instead.   I added a makefile target to run codespell locally and cleaned up the reported errors.
- I added a very basic `yaml` linter to the GH actions and a Makefile target to run it.  The next step is using our own linkml-lint to do the equivalent level of linting for these older schemas that might take awhile to move to the new standards (e.g. Biolink, linkml-model, etc.).
   -  while in the GH actions, I also added a bit of code to re-run the tests on modification of a PR. 
- I made `codespell`, `yamllint` dev dependencies in `pyproject.toml`.  I know there is more to do in the pyproject.toml file - I stopped short of doing the rest!
- It's unclear to me why we need `setup.py` and `setup.cfg` any longer.  These are also removed in this PR.
- finally, I added `linkml:notes` to all our Types classes to indicate on our generated doc, that we should use lowercase names for these instead of uppercase.


.github/workflows/main.yaml
```diff
@@ -4,7 +4,10 @@
 
 name: Build and test linkml-model
 
-on: [pull_request]
+on:
+  pull_request:
+    branches: [ main ]
+    types: [ opened, synchronize, reopened ]
```
pkalita-lbl, line 10: I think it's fine to explicitly specify the activity types here, but just FYI these are the default anyway:

> By default, a workflow only runs when a `pull_request` event's activity type is `opened`, `synchronize`, or `reopened`. To trigger workflows by different activity types, use the types keyword. 

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

.github/workflows/validate-yaml-syntax.yaml
```diff
@@ -0,0 +1,15 @@
...
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+jobs:
+  validate-yaml:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Validate YAML file
+        run: yamllint -c .yamllint-config linkml_model/model/schema/*.yaml
```
pkalita-lbl, line 15: How does `yamllint` get installed in this workflow?

Makefile
```diff
@@ -122,4 +122,21 @@ clean:
...
+		base=$$(basename $$file); \
+		filename_without_suffix=$${base%.*}; \
+		poetry run gen-python --genmeta $$file > $(PYMODEL)/$$filename_without_suffix.py; \
+	done
+
+spell:
+	poetry run codespell
+
+lint:
+	poetry run yamllint -c .yamllint-config linkml_model/model/schema/*.yaml
```
pkalita-lbl, line 139: What do you think about using one of the three standard config file names (`.yamllint`, `.yamllint.yaml`, or `.yamllint.yml` -- https://yamllint.readthedocs.io/en/stable/configuration.html) so that the `-c` flag doesn't have to be specified every time?

linkml_model/model/docs/specification/03schemas.md
```diff
@@ -72,22 +72,22 @@ However, it should be understood these are not precisely equivalent.
...
-| [classes](classes.md) | 0..* <br/> [ClassDefinition](ClassDefinition.md)  | An index to the collection of all class definitions in the schema   |
-| [slot_definitions](slot_definitions.md) | 0..* <br/> [SlotDefinition](SlotDefinition.md)  | An index to the collection of all slot definitions in the schema   |
-| [enums](enums.md) | 0..* <br/> [EnumDefinition](EnumDefinition.md)  | An index to the collection of all enum definitions in the schema   |
-| [subsets](subsets.md) | 0..* <br/> [SubsetDefinition](SubsetDefinition.md)  | An index to the collection of all subset definitions in the schema   |
-| [prefixes](prefixes.md) | 0..* <br/> [Prefix](Prefix.md)  | prefix / URI definitions to be added to the context beyond those fetched from...   |
-| [default_prefix](default_prefix.md) | 0..1 <br/> [xsd:string](http://www.w3.org/2001/XMLSchema#string)  | default and base prefix -- used for ':' identifiers, @base and @vocab   |
-| [default_range](default_range.md) | 0..1 <br/> [TypeDefinition](TypeDefinition.md)  | default slot range to be used if range element is omitted from a slot definit...   |
```
pkalita-lbl, line 85: This is probably tangential to your changes, but what's up with all these tables with truncated lines? Are these autogenerated from something?

pkalita-lbl (commented):
Hahah woops, a couple minutes late on this review 😂 But just a couple of minor comments.

cmungall:
yep, those lines are autogenerated. Maybe better just to link out.

.github/workflows/validate-yaml-syntax.yaml
sierra-moxon: yamllint is already installed on Ubuntu-based GitHub runners.

pkalita-lbl: Whoa! https://github.com/actions/runner-images/issues/1142 So it is. The more you know!

PR #168, opened 2023-08-30
Use versioned documentation deployment

pkalita-lbl:
Fixes https://github.com/linkml/linkml/issues/1590

These changes use [mike](https://github.com/jimporter/mike) to have multiple versions of the documentation deployed. Changes to the main branch will be deployed as a "dev" version. Each release is deployed as a "MAJOR.MINOR" version -- that is, releasing 1.5.3 will replace the 1.5.2 docs. But releasing 1.6.0 will not affect the 1.5.3 docs. The latest version will also be accessible via the "latest" alias. See https://pkalita-lbl.github.io/linkml-model for a demo of how it works with a bunch of artificial releases.

Things to note:
* Since GitHub Pages doesn't offer any server-level mechanism to do redirecting or URL rewriting, we have to be a little clever to get URLs like `https://linkml.io/linkml-model/docs/pattern` to redirect to `https://linkml.io/linkml-model/latest/docs/pattern`. It is accomplished with the `404.html` file here. It needs to live outside of any of the versioned doc builds, so that's why you'll see the workflows manually copying this file.
* When a release is made an extra copy of the `linkml_model` directory is also made at the root-level of the `gh-pages` branch so that URLs like `https://linkml.io/linkml-model/linkml_model/model/schema/types.yaml` continue to work, in addition to new versioned URLs like `https://linkml.io/linkml-model/1.6.0/linkml_model/model/schema/types.yaml`
* The `main` branch used to have generated docs in the `docs` directory. To remove any confusion I've deleted all those files, and the `docs` directory is now git-ignored. Doc files are built as they are needed both locally and in workflows so there's no need to check in these files.
* Renamed a couple `Makefile` targets for consistency (`gendoc -> gen-doc`, `generate_python_models -> gen-py`)
* Removed extraneous `Pipfile`, `requirements.txt`, `requirements-dev.txt`

## 🚨 WARNING 🚨 
Because this completely restructures the `gh-pages` branch we should only merge this when we're ready to make a new release. The process should be:
* Manually clear out the existing `gh-pages` branch
* Merge this PR. The triggered action will set up the "dev" version of the docs
* Make a new release. The triggered action will set up the first numbered version of docs and the "latest" alias. 

I'm marking this PR as a draft just so it doesn't get accidentally merged at the wrong time, but I don't anticipate making further changes.


Makefile
```diff
@@ -11,9 +11,15 @@ RUN = poetry run
 SCHEMA_NAME = linkml_model
 SOURCE_SCHEMA_PATH = $(shell sh ./utils/get-value.sh source_schema_path)
 SRC = .
-DEST = target
-PYMODEL = linkml_model/
-DOCDIR = $(DEST)/docs
+DEST = staging
+PYMODEL = linkml_model
+DOCDIR = docs/source
```
sierra-moxon, line 16: DOCDIR changes b/c of `mike` requirements?

sierra-moxon (approved):

Makefile
pkalita-lbl: Sort of. Conceptually the "documentation" for `linkml-model` (i.e. what's accessible from `https://linkml.io/linkml-model`) needs to comprise the actual generated doc pages (served under the `/docs` route), the YAML source files (like `types.yaml`) and other generated artifacts (like `types.shex`). Currently we accomplish that by an unholy combination of running mkdocs _locally_ and GitHub Pages' automatic Jekyll configuration (tied to the `main` branch). But, yeah, `mike` needs to be part of a strictly mkdocs-based workflow. So that means we need to wrangle all the necessary files into place and then let `mkdocs`/`mike` process them. That's what `DOCDIR` is now.

sujaypatil96 (approved):

PR #169, opened 2023-09-01
Fix repository name in release workflow

pkalita-lbl:


pkalita-lbl:
### 🤦🏻 

It looks like now that we've released v1.6.0 we have more releases than [this request](https://github.com/linkml/linkml-model/blob/55037f041ecf9e05e0df80dcb82f8d72b567d8fc/linkml_model/linkml_files.py#L114) gets in one shot. We could probably kick the can down the road a bit by adding [adding a per_page query parameter](https://docs.github.com/en/rest/guides/using-pagination-in-the-rest-api?apiVersion=2022-11-28#changing-the-number-of-items-per-page). Or we could make that method smarter by following pagination links until it finds what it's looking for or until it runs out of pages.

pkalita-lbl:
Since I'm going to be out of the office for a bit I've got ahead and kicked the can (i.e. using the `per_page` parameter) so that I'm not leaving this in a broken state. I can revisit it when I'm back if we think a more robust solution is needed.

cmungall (approved):

PR #170, opened 2023-09-14
Remove Mark from author list

nlharris:
Mark asked to be removed (he probably doesn't want his personal email address shared). It's not clear who should be added. So for now, let's remove Mark and just leave Chris. fixes #1142


turbomam (approved):
I didn't really care about expressing my email, I just didn't think I deserved credit for contributing to the metamodel

PR #172, opened 2023-11-17
XY alias removal of axes

melonora:
closes [#1740](https://github.com/linkml/linkml/issues/1740)

As described in the issue I think the x y aliases for axis 0 and 1 should be removed as this very much depends on the array implementation.


rly:
This sounds reasonable to me.

PR #176, opened 2024-01-15
Make maximum_value and minimum_value accept linkml:Any

sneakers-the-rat:
Fix: https://github.com/linkml/linkml/issues/1384

Per the discussion on 1384, Relax the range constraint on `minimum_value` and `maximum_value` to accept `linkml:Any`, pending a fuller rework to create an `Ordinal` meta-type. 

Changes
- Updated `range: Anything`
- Updated slot descriptions to anticipate ordinal type - I figured this would be easier than trying to say "For now you can put anything here, but in the future..." 
- Add `notes` to remind about future plans
- Update docs to match.

I tried running `gen-project` but it seemed to not just change the `minimum_value` and `maximum_value`, but rework most of the generated schema, so I held off adding that to the PR - not sure how that typically works in PRs for this repo. lmk if i should include.

### DIFF
--- a/linkml_model/model/docs/specification/03schemas.md
+++ b/linkml_model/model/docs/specification/03schemas.md
@@ -352,8 +352,8 @@ Any instance *s* of a SlotDefinition may have assignments in any of the followin
 | [range](../range.md) | 0..1 <br/> [Element](../Element.md)  | defines the type of the object of the slot   |
 | [range_expression](../range_expression.md) | 0..1 <br/> [AnonymousClassExpression](../AnonymousClassExpression.md)  | A range that is described as a boolean expression combining existing ranges  |
 | [enum_range](../enum_range.md) | 0..1 <br/> [EnumExpression](../EnumExpression.md)  | An inlined enumeration   |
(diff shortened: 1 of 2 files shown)


cmungall (approved):

turbomam:
The ranges asserting here are `Anything`. Isn't there also a LinkML `Any`? What's the difference?

sneakers-the-rat:
I just followed the example of the other existing Anything range. It looks like its an alias:
https://github.com/sneakers-the-rat/linkml-model/blob/3ad48fc0c1fd0cc5664ac85f36b26aea7c7901c2/linkml_model/model/schema/meta.yaml#L2345

Docs have an Anything page: https://linkml.io/linkml-model/latest/docs/Anything/

But no Any: https://linkml.io/linkml-model/latest/docs/Any

Accordingly W3ID purl for Any does not dereference: https://w3id.org/linkml/Any

But Anything does:
https://w3id.org/linkml/Anything

The mappings on that page are similarly ambiguous. So it seems like its an alias but an incomplete/imperfect one?

ddooley:
Now I presume the LinkML validation scripts still continue with business as usual by testing integers against minimum_value and maximum_value, but now LinkML does not apply that validation to anything other than integers?

PR #181, opened 2024-02-05
First pass at native NDArray support.

cmungall:
See

- https://github.com/linkml/linkml/issues/895
- https://github.com/linkml/linkml-arrays/issues/5

This introduces first-class array support into LinkML. 

A minimal example would be:

```yaml
    attributes:
      temperature_matrix:
        range: float
        array_info:
          exact_dimensions: 3
```

The native serialization of this in json/yaml will be a LoLoL. Using linkml-xarrays it will be possible to serialize using hdf5/zarr/etc.

The corresponding nptyping type would be `NDArray[Shape["*, *, *"], Float]`.

(note: modelers will want the ability to use ctypes but this is orthogonal)

Note that this does not force any metadata on the array; we are deferring on the datamodel for what is equivalent to xarray DataArrays, these will be supported via `implements` for now and first-class incorporation in a future version. This will allow binding between axes are other LinkML arrays.

Minimal metadata can be introduced via naming the axes

```yaml
    attributes:
      temperature_matrix:
        range: float
        array_info:
          exact_dimensions: 3
          dimensions:
            x:
            y:
            z:
```

The corresponding nptyping type would be `NDArray[Shape["* x, * y, * z"], Float]`.

The shape can be further constrained; imagine an RGB matrix with coords x, y, and a length 3 r/g/b:

```yaml
    attributes:
      rgb:
        range: float
        array_info:
          exact_dimensions: 3
          dimensions:
            x:
            y:
            rgb:
              exact_cardinality: 3
              description: r, g, b values
              annotations:
                names: "[red, green, blue]"
```

corresponds to `NDArray[Shape["* x, * y, 3 rgb"]`

For now if you do want to bind dimensions to additional metadata this can be done via annotations:

```yaml
classes:

  TemperatureDataset:
    tree_root: true
    annotations:
      array_data_mapping:
        data: temperatures_in_K
        dims: [x, y, t]
        coords:
          latitude_in_deg: x
          longitude_in_deg: y
          time_in_d: t
    attributes:
      name:
        identifier: true
        range: string
      latitude_in_deg:
        required: true
        range: float
        multivalued: true
        unit:
          ucum_code: deg
        array_info:
          exact_dimensions: 1
      longitude_in_deg:
        required: true
        range: float
        multivalued: true
        unit:
          ucum_code: deg
        array_info:
          exact_dimensions: 1
      time_in_d:
        range: float
        multivalued: true
        implements:
          - linkml:elements
        required: true
        unit:
          ucum_code: d
        array_info:
          exact_dimensions: 1
      temperatures_in_K:
        range: float
        multivalued: true
        required: true
        unit:
          ucum_code: K
        array_info:
          exact_dimensions: 3
```

### DIFF
--- a/linkml_model/model/schema/meta.yaml
+++ b/linkml_model/model/schema/meta.yaml
@@ -1570,7 +1570,7 @@ slots:
     inherited: true
     description: the exact number of entries for a multivalued slot
     in_subset:
-      - SpecificationSubse
+      - SpecificationSubset
     comments:
       - if exact_cardinality is set, then minimum_cardinalty and maximum_cardinality must be unset or have the same value
 
--- a/linkml_model/model/schema/meta.yaml
+++ b/linkml_model/model/schema/meta.yaml
@@ -1427,10 +1427,10 @@ slots:
     description: coerces the value of the slot into an array and defines the dimensions of that array
     status: testing
 
-  array_axes:
+  dimensions:
     description: definitions of each axis in the array
     domain: array_info_expression
-    range: array_axis_expression
+    range: dimension_expression
     multivalued: true
     inlined: true
     status: testing
@@ -2970,10 +2970,10 @@ classes:
       - exact_dimensions
       - minimum_dimensions
       - maximum_dimensions
-      - array_axes
+      - dimensions
     status: testing
 
-  array_axis_expression:
+  dimension_expression:
     description: defines one of the dimensions of an array
     mixins:
       - extensible
--- a/linkml_model/model/schema/meta.yaml
+++ b/linkml_model/model/schema/meta.yaml
@@ -1419,36 +1419,48 @@ slots:
       - BasicSubset
       - ObjectOrientedProfile
 
-  array_info:
+  array:
     domain: slot_definition
-    range: array_info_expression
+    range: array_expression
     inherited: true
     description: coerces the value of the slot into an array and defines the dimensions of that array
     status: testing
 
-  dimensions_info:
+  dimensions:
+    aliases:
+      - dimensions
     description: definitions of each axis in the array
-    domain: array_info_expression
+    domain: array_expression
     range: dimension_expression
     multivalued: true
+    list_elements_ordered: true
     status: testing
 
   minimum_dimensions:
     description: minimum number of dimensions in the array
-    domain: array_info_expression
+    domain: array_expression
     range: integer
     status: testing
+    minimum_value: 1
 
   maximum_dimensions:
     description: maximum number of dimensions in the array
-    domain: array_info_expression
+    domain: array_expression
     range: integer
     status: testing
+    minimum_value: 1
 
   exact_dimensions:
     description: exact number of dimensions in the array
-    domain: array_info_expression
+    domain: array_expression
     range: integer
+    status: testing
+    minimum_value: 1
+
+  has_extra_dimensions:
+    description: If this is set to true
+    domain: array_expression
+    range: boolean
     status: testing
 
   inherited:
@@ -1558,7 +1570,9 @@ slots:
     inherited: true
     description: the exact number of entries for a multivalued slot
     in_subset:
-      - SpecificationSubset
+      - SpecificationSubse
+    comments:
+      - if exact_cardinality is set, then minimum_cardinalty and maximum_cardinality must be unset or have the same value
 
   minimum_cardinality:
     is_a: list_value_specification_constant
@@ -1567,6 +1581,7 @@ slots:
     description: the minimum number of entries for a multivalued slot
     in_subset:
       - SpecificationSubset
+    minimum_value: 0
 
   maximum_cardinality:
     is_a: list_value_specification_constant
@@ -1575,6 +1590,9 @@ slots:
     description: the maximum number of entries for a multivalued slot
     in_subset:
       - SpecificationSubset
+    comments:
+      - maximum_cardinality cannot be less than minimum_cardinality
+    minimum_value: 0
 
   equals_string_in:
     is_a: list_value_specification_constant
@@ -2799,7 +2817,7 @@ classes:
       - domain
       - slot_uri
       - multivalued
-      - array_info
+      - array
       - inherited
       - readonly
       - ifabsent
@@ -2958,7 +2976,7 @@ classes:
     in_subset:
       - SpecificationSubset
 
-  array_info_expression:
+  array_expression:
     description: defines the dimensions of an array
     mixins:
       - extensible
@@ -2968,7 +2986,8 @@ classes:
       - exact_dimensions
       - minimum_dimensions
       - maximum_dimensions
-      - dimensions_info
+      - has_extra_dimensions
+      - dimensions
     status: testing
 
   dimension_expression:
--- a/linkml_model/model/schema/meta.yaml
+++ b/linkml_model/model/schema/meta.yaml
@@ -1442,6 +1442,8 @@ slots:
     range: integer
     status: testing
     # minimum_value: 1
+    comments:
+      - minimum_cardinality cannot be greater than maximum_cardinality
 
   maximum_number_dimensions:
     description: maximum number of dimensions in the array, or False if explicitly no maximum.
@@ -1464,6 +1466,8 @@ slots:
     range: integer
     status: testing
     # minimum_value: 1
+    comments:
+      - if exact_number_dimensions is set, then minimum_number_dimensions and maximum_number_dimensions must be unset or have the same value
 
   has_extra_dimensions:
     description: If this is set to true
@@ -1590,6 +1594,8 @@ slots:
     in_subset:
       - SpecificationSubset
     # minimum_value: 0
+    comments:
+      - minimum_cardinality cannot be greater than maximum_cardinality
 
   maximum_cardinality:
     is_a: list_value_specification_constant
--- a/linkml_model/model/schema/meta.yaml
+++ b/linkml_model/model/schema/meta.yaml
@@ -1420,6 +1420,34 @@ slots:
       - BasicSubset
       - ObjectOrientedProfile
 
+  array_shape:
+    domain: slot_definition
+    range: array_shape_expression
+    inherited: true
+    description: coerces the value of the slot into an array and defines the dimensions of that array
+
+  array_axes:
+    description: definitions of each axis in the array
+    domain: array_shape_expression
+    range: array_axis_expression
+    multivalued: true
+    inlined: true
+
+  minimum_dimensions:
+    description: minimum number of dimensions in the array
+    domain: array_shape_expression
+    range: integer
+
+  maximum_dimensions:
+    description: maximum number of dimensions in the array
+    domain: array_shape_expression
+    range: integer
+
+  exact_dimensions:
+    description: exact number of dimensions in the array
+    domain: array_shape_expression
+    range: integer
+
   inherited:
     domain: slot_definition
     range: boolean
@@ -1518,6 +1546,14 @@ slots:
     see_also:
       - https://linkml.io/linkml/developers/inference.html
       - https://linkml.io/linkml/schemas/advanced.html#equals-expression
+    in_subset:
+      - SpecificationSubset
+
+  exact_cardinality:
+    is_a: list_value_specification_constant
+    range: integer
+    inherited: true
+    description: the exact number of entries for a multivalued slot
     in_subset:
       - SpecificationSubset
 
@@ -2715,6 +2751,7 @@ classes:
       - equals_string_in
       - equals_number
       - equals_expression
+      - exact_cardinality
       - minimum_cardinality
       - maximum_cardinality
       - has_member
@@ -2759,6 +2796,7 @@ classes:
       - domain
       - slot_uri
       - multivalued
+      - array_shape
       - inherited
       - readonly
       - ifabsent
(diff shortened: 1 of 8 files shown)


tests/input/examples/schema_definition-native-array-rgb.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+imports:
+  - linkml:types
+
+classes:
+
+  RGBImage:
+    attributes:
+      rgb:
+        range: float
```
rly, line 23:
```suggestion
        range: float
        multivalued: true
```

Is this necessary?

tests/input/examples/schema_definition-native-array-rgb.yaml
```diff
@@ -0,0 +1,44 @@
...
+            x:
+            y:
+            rgb:
+              exact_cardinality: 3
+              description: r, g, b values
+              annotations:
+                names: "[red, green, blue]"
+  SquareDataset:
+    attributes:
+      square_matrix:
```
rly, line 37:
```suggestion
      square_matrix:
        multivalued: true
```

Same here.

tests/input/examples/schema_definition-array-2.yaml
```diff
@@ -26,7 +26,7 @@ classes:
     implements:
       - linkml:NDArray
     annotations:
-      dimensions: 1
+      dimensions_info: "1"
```
rly, line 29: Should I reformat these to use:
```yaml
        array_info:
          exact_dimensions: 1
```

rly: For this TimestampSeries class, I'm actually not sure it would be allowed to put `array_info` at the class definition level, but maybe for the other ones?

tests/input/examples/schema_definition-native-array-rgb.yaml
```diff
@@ -0,0 +1,44 @@
...
+          # NPtyping: NDArray[Shape["* x, * y, 3 rgb"]
+          exact_dimensions: 3
+          dimensions_info:
+            x:
+            y:
+            rgb:
+              exact_cardinality: 3
+              description: r, g, b values
+              annotations:
+                names: "[red, green, blue]"
```
rly, line 34:
```suggestion
            x:
              dimension_index: 0
            y:
              dimension_index: 1
            rgb:
              dimension_index: 2
              exact_cardinality: 3
              description: r, g, b values
              annotations:
                names: "[red, green, blue]"
```

It would be good to be explicit here. This also allows some dimensions to be unnamed.

rly:
Notes:
- It is possible to have scalar arrays in many array formats (e.g., NumPy, HDF5, Zarr) but not all (e.g., Matlab). They seem primarily used to allow API functions to work seamlessly with both arrays and scalars and as a box to pass scalars to a function by reference. Matlab gets around this by not having special dtypes for scalars and not allowing scalar arrays - they are just 1x1 matrices. I do not know that we need to support scalar arrays. To be conservative, perhaps it would be best to leave out support until there is a need. Refs:
  - [Interesting comment on why scalar arrays exist in numpy](https://stackoverflow.com/a/794812/20177)
  - [Description of scalar dataspaces in HDF5](https://davis.lbl.gov/Manuals/HDF5-1.8.7/UG/UG_frame12Dataspaces.html) 
  - [Use of 0-D arrays as a data structure in Julia](https://discourse.julialang.org/t/why-have-0-dimensional-arrays/797)
  - [Use of 0-D arrays as a data structure in Lisp](https://stackoverflow.com/questions/77078104/what-are-the-use-cases-for-zero-dimensions-array-in-common-lisp)
- If not specified, `minimum_number_dimensions` is 1, `maximum_number_dimensions` is Infinity.
- It is possible to specify `minimum_number_dimensions` without `maximum_number_dimensions`. We have a use case for a min without a max, but I cannot think of one for a max without a min.
- I am not sure there is a use case for `minimum_cardinality` and `maximum_cardinality`. Maybe we can omit them and add them later if there is a need? I am concerned about adding features to the metamodel that are useless but need to be supported for a long time to avoid breaking backwards compatibility.

Some useful validation checks:
1. The number of items in `dimensions_info` must not exceed `maximum_number_of_dimensions`
2. Each `dimension_index` must not exceed `maximum_number_of_dimensions - 1`
3. Each `dimension_index` must be unique across `dimensions_info` items
4. If `array_info` exists, then `multivalued` must be True (if we allow scalar arrays, then this is no longer always true)
7. `minimum_number_dimensions` <= `maximum_number_dimensions`
8. Cannot have both `exact_number_dimensions` and `minimum_number_dimensions` or both `exact_dimensions` and `maximum_number_dimensions`
9. `minimum_number_dimensions`, `exact_number_dimensions`, `maximum_number_dimensions` all > 0
10. `minimum_cardinality` <= `maximum_cardinality`
5. Cannot have both `exact_cardinality` and `minimum_cardinality` or both `exact_cardinality` and `maximum_cardinality`
6. `minimum_cardinality`, `exact_cardinality`, `maximum_cardinality` all > 0

tests/input/examples/schema_definition-native-array-rgb.yaml
cmungall: multivalued will be implicit based on the presence of `array_info`. There are a few other cases like that e.g. `inlined_as_dict -> inlined`

tests/input/examples/schema_definition-array-2.yaml
cmungall: unfortunately that doesn't quite work because the range of `dimensions_info` is a collection of `dimension_expression`s, so these have to be at a different level

e.g. this is invalid yaml:

```yaml
dimensions_info: 3
  x:
  y:
  z:
```

linkml_model/model/schema/meta.yaml
```diff
@@ -1420,6 +1420,39 @@ slots:
...
       - ObjectOrientedProfile
 
+  array_info:
+    domain: slot_definition
+    range: array_info_expression
+    inherited: true
+    description: coerces the value of the slot into an array and defines the dimensions of that array
+    status: testing
+
+  dimensions_info:
```
sneakers-the-rat, line 1430: hopefully not annoying naming question, why not just `array` and `dimensions` here?

linkml_model/model/schema/meta.yaml
```diff
@@ -1420,6 +1420,39 @@ slots:
...
+    description: definitions of each axis in the array
+    domain: array_info_expression
+    range: dimension_expression
+    multivalued: true
+    inlined: true
+    status: testing
+
+  minimum_dimensions:
```
sneakers-the-rat, line 1438: how do we model mutual exclusivity and constraints between properties here? eg. one shouldn't use `exact_dimensions` alongside `maximum_dimensions`, `maximum_dimensions` shouldn't be less than `minimum_dimensions`, etc. Part of why i like expressing ranges/values in a single object like `dimensions: 3` or `dimensions: {min: 2, max: 3}` is that you can model within the scope of that object, but it seems like that can also be done here i'm just not sure how

linkml_model/model/schema/meta.yaml
```diff
@@ -1420,6 +1420,39 @@ slots:
...
+    description: minimum number of dimensions in the array
+    domain: array_info_expression
+    range: integer
+    status: testing
+
+  maximum_dimensions:
```
sneakers-the-rat, line 1444: separate but related question to above - it seems like ranges and exact values will/is a common pattern, thoughts on having a range syntax so we could just have a single property that can take an integer or a `1..2` specification?

linkml_model/model/schema/meta.yaml
```diff
@@ -1420,6 +1420,39 @@ slots:
       - BasicSubset
       - ObjectOrientedProfile
 
+  array_info:
```
sneakers-the-rat, line 1423: are there no minimum properties that need to be specified on an array? eg. having this property without any values is an `Any` shaped array?

tests/input/examples/schema_definition-native-array-rgb.yaml
```diff
@@ -0,0 +1,44 @@
...
+  SquareDataset:
```
sneakers-the-rat, line 35: specifically on this idea of expressing a square dataset, where we want to specify "array can be any cardinality where the cardinality of `x` and `y` are equal, would it be possible to specify cardinality and dimension as an expression like `exact_dimensions: "prefix:Dimension * 2"` or something like that? or maybe that's a v2 kind of thing

sneakers-the-rat (commented):
OK I am mostly settled on this, but I have a pitch that i'll make in person about extending the notion of `implements` and how it could allow for plugins in a minute (and then update this so that it's public later)

tests/input/examples/schema_definition-native-array-rgb.yaml
rly: discussed - we do not want to make a DSL right now

linkml_model/model/schema/meta.yaml
sneakers-the-rat: (we did change it to that)

linkml_model/model/schema/meta.yaml
```diff
@@ -1420,6 +1419,50 @@ slots:
...
+  array:
+    domain: slot_definition
+    range: array_expression
+    inherited: true
+    description: coerces the value of the slot into an array and defines the dimensions of that array
+    status: testing
+
+  dimensions:
+    aliases:
+      - dimensions
```
rly, line 1431:
```suggestion
```

rly, line 1431:
```suggestion
    aliases:
      - axes
```

linkml_model/model/schema/meta.yaml
```diff
@@ -1521,13 +1564,24 @@ slots:
     in_subset:
       - SpecificationSubset
 
+  exact_cardinality:
+    is_a: list_value_specification_constant
+    range: integer
+    inherited: true
+    description: the exact number of entries for a multivalued slot
+    in_subset:
+      - SpecificationSubse
```
rly, line 1573:
```suggestion
      - SpecificationSubset
```

linkml_model/model/schema/meta.yaml
```diff
@@ -1521,13 +1564,24 @@ slots:
...
+    comments:
+      - if exact_cardinality is set, then minimum_cardinalty and maximum_cardinality must be unset or have the same value
+
```
rly, line 1576:
```suggestion
    minimum_value: 0

```

linkml_model/model/schema/meta.yaml
```diff
@@ -1521,13 +1564,24 @@ slots:
...
   minimum_cardinality:
     is_a: list_value_specification_constant
     range: integer
     inherited: true
     description: the minimum number of entries for a multivalued slot
     in_subset:
       - SpecificationSubset
+    minimum_value: 0
```
rly, line 1584:
```suggestion
    comments:
      - minimum_value cannot be greater than maximum_cardinality
    minimum_value: 0
```

linkml_model/model/schema/meta.yaml
```diff
@@ -1420,6 +1419,50 @@ slots:
...
+    range: integer
+    status: testing
+    minimum_value: 1
+
+  exact_number_dimensions:
+    description: exact number of dimensions in the array
+    domain: array_expression
+    range: integer
+    status: testing
+    minimum_value: 1
```
rly, line 1458:
```suggestion
    comments:
      - if exact_number_dimensions is set, then minimum_number_dimensions and maximum_number_dimensions must be unset or have the same value
    minimum_value: 1
```

linkml_model/model/schema/meta.yaml
```diff
@@ -1420,6 +1419,50 @@ slots:
...
+    multivalued: true
+    list_elements_ordered: true
+    status: testing
+
+  minimum_number_dimensions:
+    description: minimum number of dimensions in the array
+    domain: array_expression
+    range: integer
+    status: testing
+    minimum_value: 1
```
rly, line 1444:
```suggestion
    comments:
    - minimum_number_dimensions cannot be greater than maximum_number_dimensions
    minimum_value: 1
```

linkml_model/model/schema/meta.yaml
```diff
@@ -1420,6 +1419,50 @@ slots:
...
+
+  maximum_number_dimensions:
+    description: maximum number of dimensions in the array
+    domain: array_expression
+    range: integer
+    status: testing
+    minimum_value: 1
```
rly, line 1451:
```suggestion
    comments:
      - maximum_number_dimensions cannot be less than minimum_number_dimensions
    minimum_value: 1
```

sneakers-the-rat:
for posterity, archived version of matrix of tradeoffs to be added to consolidated docs later, along with @rly 's notes and examples :) 

live: https://wiki.jon-e.net/LinkML_Arrays
archive: https://web.archive.org/save/https://wiki.jon-e.net/LinkML_Arrays

linkml_model/model/schema/meta.yaml
```diff
@@ -2917,6 +2976,34 @@ classes:
...
+
+  dimension_expression:
+    description: defines one of the dimensions of an array
+    mixins:
+      - extensible
+      - annotatable
+      - common_metadata
+    slots:
+      - alias
+      - equals_expression
```
rly, line 3001:
```suggestion
```
Perhaps remove?

cmungall: Done!

linkml_model/model/schema/meta.yaml
```diff
@@ -1420,6 +1419,58 @@ slots:
...
+    range: integer
+    status: testing
+    # minimum_value: 1
+
+  has_extra_dimensions:
+    description: If this is set to true
+    domain: array_expression
+    range: boolean
+    status: testing
+
```
rly, line 1473:
```suggestion
```

linkml_model/model/schema/meta.yaml
```diff
@@ -1420,6 +1419,62 @@ slots:
...
+    # minimum_value: 1
+    comments:
+      - if exact_number_dimensions is set, then minimum_number_dimensions and maximum_number_dimensions must be unset or have the same value
+
+  has_extra_dimensions:
+    description: If this is set to true
+    domain: array_expression
+    range: boolean
+    status: testing
+
```
rly, line 1477:
```suggestion
```

linkml_model/model/schema/meta.yaml
```diff
@@ -2917,6 +2990,33 @@ classes:
...
+    description: defines the dimensions of an array
+    mixins:
+      - extensible
+      - annotatable
+      - common_metadata
+    slots:
+      - exact_number_dimensions
+      - minimum_number_dimensions
+      - maximum_number_dimensions
+      - has_extra_dimensions
```
rly, line 3003:
```suggestion
```

PR #182, opened 2024-02-07
Mention  BasicSubset just once

yarikoptic:


sierra-moxon:
Thanks @yarikoptic - we should make the change in the [yaml](https://github.com/linkml/linkml-model/blob/aab9842be0e230c0040688dfc6ffa26696c97827/linkml_model/model/schema/meta.yaml#L26) file, as this python is autogenerated.

sierra-moxon (changes requested):
remove this duplication from the source yaml vs. the autogenerated python here.

sierra-moxon (approved):

PR #186, opened 2024-02-10
Remove Any range in maximum_number_dimensions

sneakers-the-rat:
Noticed that `ArrayExpression.maximum_number_dimensions` had the wrong type: https://github.com/linkml/linkml-runtime/blob/c6943e41b67c9eb24f056c6ebaea2c78a0bdf4af/linkml_runtime/linkml_model/meta.py#L2730

```python
maximum_number_dimensions: Optional[Union[dict, Anything]] = None
```

should be

```python
maximum_number_dimensions: Optional[Union[int, bool]] = None
```

looks like a `range: Anything` was left in, I'm not sure if that resolves the range generation, but it also seems like something that should be validated in the metamodel schema: `range` shouldn't be specified by itself and in an `any_of` expression

### DIFF
--- a/linkml_model/model/schema/meta.yaml
+++ b/linkml_model/model/schema/meta.yaml
@@ -1457,7 +1457,6 @@ slots:
       as a closed list and the maximum_number_dimensions is the length of the dimensions list, unless this
       value is set to False
     domain: array_expression
-    range: Anything
     any_of:
       - range: integer
         # minimum_value: 1


cmungall:
I think this needs to be left in for safety just now, to avoid non-monotonicity. The best fix here is in `gen-python` which isn't yet on a par with gen-pydantic for generating unions

sneakers-the-rat:
Not sure what you mean here, is it valid syntax to have

```yaml
range: integer
any_of: 
  - range: Anything
```

I would have figured that the syntax would be 

```yaml
any_of:
  - range: integer
  - range: Anything
```

in any case i'm not sure why `dict` is showing up there. I could take a look the dataclass generator, so if that's the correct syntax, then i could rename this issue to something like "dataclass generator incorrectly handling range and any_of: range"

cmungall:
Should be fixed in pythongen: https://github.com/linkml/linkml/issues/1813

rly:
Resolved in @linkml/ndarray-wg discussion:
`range: Anything` is overly general on purpose but the `any_of` ranges will restrict the range of `maximum_number_dimensions`. `range: Anything` is useful for monotonicity. 

The python dataclass generator should be fixed.

PR #190, opened 2024-03-10
Add array class example

rly:
This PR adds an alternate approach to specifying the `TemperatureDataset` defined in `tests/input/examples/schema_definition-native-array-1.yaml`. This approach uses classes that implement [`linkml:NDArray`](https://github.com/linkml/linkml-model/blob/main/linkml_model/model/schema/array.yaml) and have an attribute that implements `linkml:elements` as defined pre-1.7.0 release. This representation is necessary for adding additional attributes, e.g., user-specified units of measurement, conversion factor, precision/error, reference/zero point, or source, on the various arrays that make up a `TemperatureDataset`. Seeking feedback on this approach.

It also changes `y` -> `"y"` because `y` = True in YAML 1.1.

### DIFF
--- a/tests/input/examples/schema_definition-native-array-2.yaml
+++ b/tests/input/examples/schema_definition-native-array-2.yaml
@@ -1,0 +1,128 @@
+id: https://example.org/arrays
+name: arrays-temperature-example-2
+title: Array Temperature Example Using NDArray Classes
+description: |-
+  Example LinkML schema to demonstrate a 3D DataArray of temperature values with labeled axes
+  using LinkML NDArray classes
+license: MIT
+
+prefixes:
+  linkml: https://w3id.org/linkml/
+  wgs84: http://www.w3.org/2003/01/geo/wgs84_pos#
+  example: https://example.org/
+
+default_prefix: example
+
+imports:
+  - linkml:types
+
+classes:
+
+  TemperatureDataset:
+    tree_root: true
+    implements:
+      - linkml:DataArray
+    annotations:
+      array_data_mapping:
+        data: temperatures_in_K
+        dims: [x, "y", t]
+        coords:
+          latitude_in_deg: x
+          longitude_in_deg: "y"
+          time_in_d: t
+    attributes:
+      name:
+        identifier: true
+        range: string
+      latitude_in_deg:
+        implements:
+          - linkml:axis  # is this necessary/useful given the above array_data_mapping?
+        range: LatitudeSeries
+        # NOTE: LatitudeSeries could have multiple array attributes. use the one that implements linkml:elements
+        required: true
+        annotations:
+          axis_index: 0  # is this necessary/useful given the above array_data_mapping?
+      longitude_in_deg:
+        implements:
+          - linkml:axis
+        range: LongitudeSeries
+        required: true
+        annotations:
+          axis_index: 1
+      time_in_d:
+        implements:
+          - linkml:axis
+        range: DaySeries
+        required: true
+        annotations:
+          axis_index: 2
+      temperatures_in_K:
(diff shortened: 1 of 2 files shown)


tests/input/examples/schema_definition-native-array-2.yaml
```diff
@@ -0,0 +1,127 @@
...
+          latitude_in_deg: x
+          longitude_in_deg: "y"
+          time_in_d: t
+    attributes:
+      name:
+        identifier: true
+        range: string
+      latitude_in_deg:
+        implements:
+          - linkml:axis
```
rly, line 39: Is this necessary anymore given the above `array_data_mapping`?

tests/input/examples/schema_definition-native-array-2.yaml
```diff
@@ -0,0 +1,127 @@
...
+        range: LatitudeSeries
+        required: true
+        annotations:
+          axis_index: 0
```
rly, line 43: Is this necessary anymore given the above `array_data_mapping`?

tests/input/examples/schema_definition-native-array-2.yaml
```diff
@@ -0,0 +1,127 @@
...
+          longitude_in_deg: "y"
+          time_in_d: t
+    attributes:
+      name:
+        identifier: true
+        range: string
+      latitude_in_deg:
+        implements:
+          - linkml:axis
+        range: LatitudeSeries
```
rly, line 40: # NOTE: `LatitudeSeries` could have multiple array attributes. Use the one that implements `linkml:elements`. We should also ensure that exactly one attribute implements `linkml:elements`.

tests/input/examples/schema_definition-native-array-2.yaml
```diff
@@ -0,0 +1,127 @@
...
+      time_in_d:
+        implements:
+          - linkml:axis
+        range: DaySeries
+        required: true
+        annotations:
+          axis_index: 2
+      temperatures_in_K:
+        implements:
+          - linkml:array
```
rly, line 60: Should we rename this in the metamodel? Is this scoped to `linkml:DataArray` or unique across LinkML?

rly:
cc @sneakers-the-rat @cmungall 

I'm looking at how to 1) add attributes to arrays and 2) support labeling arrays with other arrays. 

For 1), I think we need to keep supporting classes that implement `linkml:NDArray` and allow those to be an axis in a `linkml:DataArray`. Alternatively, we could add attributes to `TemperatureDataset` and have them share a common prefix, e.g., `latitude_in_deg__precision`, but this is kinda ugly and relies on a naming convention. Note that numpy arrays and other simple array formats do not allow attributes, but HDF5 (and netCDF4) datasets and Zarr arrays do. Attributes are also allowed in `xarray.DataArray`.

For 2) if we allow a `linkml:NDArray` to be a labeled dimension of a `linkml:DataArray`, then because the `NDArray` class could contain multiple arrays, we need a way to identify the intended array within the class. We can keep doing that with `linkml:elements`. Alternatively, we could make `values` a special slot name for any class that implements `linkml:NDArray`. Or change the slot name in the example to `elements`. Or change `linkml:elements` to `linkml:values`. (In NWB, the Data class for arrays defines a required `data` field.)

sneakers-the-rat:
ah yes, is it time for the second leg, the indexed, array spec?

Can we get a few examples of the desired datasets we want to support with this? I think it might be helpful to have a few concrete test cases here so we can get a handle on the constraints we'll need to handle. are the ones in `linkml-arrays` still current? From that we can generate a set of requirements and constraints that help inform these decisions ^ :)

PR #194, opened 2024-05-03
Allowing permissible_value to have implements and instantiates

cmungall:
See

https://github.com/linkml/linkml/issues/2090


turbomam:
Thanks @cmungall 

Something I find challenging to keep up with: some of the "things" that determine how a generators behave are
- cli or python method arguments
- hints within the schema

If I want ed to help centralize that kind of documentation, what's the best place to put it? I probably wouldn't do it for many generators besides OWL to start

Issue #199, opened 2024-07-03
`any_of` should accept a list of `ArrayExpression`s


sneakers-the-rat:
In translating NWB, I am replacing my ArrayLike class array representations with the linkml array spec. NWB models support a `dims` and `shape` list of list expression such that a given dataset can have several different possible shapes: https://schema-language.readthedocs.io/en/stable/specification_language_description.html#dims

Given the implementation of `array` at a slot level rather than a type, I need to be able to use it in `any_of` like this:

```yaml
classes:
  MyClass:
    attributes:
      name: my_array
      range: integer
      any_of:
      - array:
          dimensions:
          - alias: x
            exact_cardinality: 3
      - array:
          dimensions:
          - alias: x
            exact_cardinality: 3
          - alias: y
            exact_cardinality: 4
```

this, of course, opens up the potential for combinatorics between `any_of` used with `range` and `array` like this:

```yaml
any_of:
- range: integer
- range: float
- array:
    dimensions:
    - alias: x
      exact_cardinality: 3
- array:
    dimensions:
    - alias: x
      exact_cardinality: 3
    - alias: y
      exact_cardinality: 4
```

which I would interpret as meaning the product:

- NDArray[Shape['3'], int]
- NDArray[Shape['3, 4'], int]
- NDArray[Shape['3'], float]
- NDArray[Shape['3, 4'], float]

and like this:

```yaml
any_of:
- range: integer
  array:
    dimensions:
    - alias: x
      exact_cardinality: 3
- range: float
  array:
    dimensions:
    - alias: x
      exact_cardinality: 3
    - alias: y
      exact_cardinality: 4
```

which would be (unambiguously)

- NDArray[Shape['3'], int]
- NDArray[Shape['3, 4'], float]

and i actually think that's a good thing - if we refactor the generators like https://github.com/linkml/linkml/pull/2019, separating the different phases of the build and using build results objects, then it would be easy to handle the combinatorics there (and i would implement it). It doesn't explode the complexity of arrays, it would just require better iteration in the generators (which is good anyway). 

without this change, I don't think there would be a way to express a slot being able to have multiple exactly parameterized/labeled array shapes. 

cc @rly

cmungall:
For your second example, each `any_of` element would be interpreted independently, and then any_of would create a union expression over them; i.e. in Python

`Union[int, float, NDArray[<R>, 3], NDArray[<R>, 4]`

The value of `<R>` would be determined by any inherited range, which may default to something like `string`

[...]

PR #200, opened 2024-07-03
move array to slot_expression

sneakers-the-rat:
fix: https://github.com/linkml/linkml-model/issues/199

I think just moving `array` to `slot_expression` would solve the problem, because `slot_definition` received the `slot_expression` mixin, which uses `slot_usage` to change the `range` of `any_of` to `anonymous_slot_expression`, which also receives the `slot_expression` mixin.

### DIFF
--- a/linkml_model/model/schema/meta.yaml
+++ b/linkml_model/model/schema/meta.yaml
@@ -2880,6 +2880,7 @@ classes:
       - exactly_one_of
       - any_of
       - all_of
+      - array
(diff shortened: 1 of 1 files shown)


cmungall (approved):

sneakers-the-rat:
bump on this, trying to close up loose ends before I can release nwb-linkml and this is one of em :)

cmungall (approved):

PR #201, opened 2024-07-15
Incorproate linkml-runtime, linkml-model/linkml-files.py changes from linkml-runtime

sierra-moxon:
please see: https://github.com/linkml/linkml-runtime/pull/320

back-porting changes in the runtime PR to linkml-model (current SoT for linkml-model/linkml_files.py methods).

this PR was a result of my dive into converting schemaview (linkml-runtime) tests to pytest when I ran into an error in linkml_files.py and test_linkml_files.py in linkml-runtime/linkml-model.  It might be that we want to reconsider vendoring. 

my biggest concern here with this PR, is that the tests for linkml-model and linkml-runtime that exercise linkml_files.py were different.  If I put back in (as I did with this PR), the tests currently in linkml-model, there are slight changes in resulting expected URLs for model files - the new URLs do resolve, so I've adjusted the tests.  We likely need @cmungall's review here before merge.


tests/test_linkml_files.py
```diff
@@ -1,64 +1,196 @@
...
-        self.assertEqual('https://raw.githubusercontent.com/linkml/linkml-model/f30637f5a585f3fc4b12fd3dbb3e7e95108d4b42/model/schema/meta.yaml', str(fileloc.meta.yaml.github_loc('v0.0.1')))
-
-
-if __name__ == '__main__':
-    unittest.main()
+    assert "https://raw.githubusercontent.com/linkml/linkml-model/f30637f5a585f3fc4b12fd3dbb3e7e95108d4b42/jsonld/meta.context.jsonld" == \
+        GITHUB_PATH_FOR(Source.META, Format.NATIVE_JSONLD, "v0.0.1")
+    current_loc = re.sub(r'linkml-model/[0-9a-f]*/', 'linkml-model/SHA/', GITHUB_PATH_FOR(Source.TYPES, Format.YAML))
+    assert "https://raw.githubusercontent.com/linkml/linkml-model/SHA/model/schema/types.yaml" == current_loc
+    assert 'https://raw.githubusercontent.com/linkml/linkml-model/missing_branch/linkml_model/owl/mappings.owl.ttl' == \
```
sierra-moxon, line 192: note: change in path here

tests/test_linkml_files.py
```diff
@@ -1,64 +1,196 @@
...
+    print(LOCAL_PATH_FOR(Source.META, Format.YAML))
+    assert os.path.join(root_path, "jsonld/types.context.jsonld") == LOCAL_PATH_FOR(Source.TYPES, Format.NATIVE_JSONLD)
+    assert "https://linkml.github.io/linkml-model/latest/linkml_model/model/schema/meta.yaml" == GITHUB_IO_PATH_FOR(
+        Source.META, Format.YAML)
+    assert "https://linkml.github.io/linkml-model/latest/linkml_model/jsonld/types.context.jsonld" == \
+           GITHUB_IO_PATH_FOR(Source.TYPES, Format.NATIVE_JSONLD)
+    assert "https://raw.githubusercontent.com/linkml/linkml-model/main/linkml_model/jsonld/meta.context.jsonld" == \
+           GITHUB_PATH_FOR(Source.META, Format.NATIVE_JSONLD, ReleaseTag.LATEST)
+    assert "https://raw.githubusercontent.com/linkml/linkml-model/testing_branch/linkml_model/owl/mappings.owl.ttl" == \
+           GITHUB_PATH_FOR(Source.MAPPINGS, Format.OWL, branch="testing_branch")
```
sneakers-the-rat, line 176: `test_basic_rules` is basically the same thing as `test_fixed_meta_url` - in general i only test fixed values when there is a specific reason for it to be fixed forever, so that's why i only tested those two URLs, but i think if we're going to test the rest of these then they should be in a test structure that reflects the meaning of the test :)

test each of these functions (`URL_FOR`, `LOCAL_PATH_FOR`, `GITHUB_IO_PATH_FOR` and `GITHUB_PATH_FOR`) separately, and some of what we want to test for might already be covered by other tests.
- `URL_FOR`, `LOCAL_PATH_FOR`, and `GITHUB_IO_PATH_FOR` are trivial string formatting functions, so they can be covered by unit-style tests. 
- `LOCAL_PATH_FOR` has a `test_local_paths` test for existence for all the files in the expected formats already.
- `URL_FOR` has `test_url_for_format` that tests the resolution of the generated URLs for the formats that have rewrite rules in the w3id repo. 
- `GITHUB_PATH_FOR` is substantially different than the others and should be tested separately for sure, 

ideally these would have docstrings about why they should be fixed forever (if they were just fixed forever, why wouldn't we just hardcode the strings in the module itself and call it a day?), but idrc about that ;)

tests/test_linkml_files.py
```diff
@@ -1,64 +1,196 @@
...
+
+
+SKIP_GITHUB_API = False
```
sneakers-the-rat, line 179: ideally this would be a pytest option or dependent on some flag, if we want to keep skipping them - i marked them as skip because they kept failing bc the ratelimits on github API were very low, but if we want to rescue them then yes a flag that can be accessed would be good :)

tests/test_linkml_files.py
sneakers-the-rat: the thing more worthy of fixing to me here is `/SHA/model` and `missing_branch/linkml_model` having different directory structure, but if the URLs resolve and they give us the thing we were after then they are the correct URLs in my book!

pyproject.toml
```diff
@@ -33,11 +33,13 @@ packages = [
 [tool.poetry.dependencies]
 python = "^3.9"
 linkml-runtime = "^1.6.3"
+rdflib = "^7.0.0"
```
sneakers-the-rat, line 36: oof if we are gaining rdflib as a runtime dependency just to use the `Namespace` class to declare a module constant i woudl call that highly undesirable and worthy of thinking about whether we need that

sneakers-the-rat (commented):
Basically i feel like i have already spent too much time for the three of us overengineering this thing, and am fine with whatever as long as the paths work ;) some minor suggestions on test semantics but take em or leave em

matentzn:
@kevinschaper can you please explore this PR and fix it or decide its no longer relevant and close it? It has a title that makes me nervous so I just want to be sure its not lost.

amc-corey-cox:
Closing as superseded — splitting the still-relevant pieces out rather than rebasing this ~2-year-old branch.

Where things landed:

- **The `Format` / `_Path` refactor** (the bulk of this PR) was independently re-implemented on `main` in #236, using a `_PathInfo` namedtuple plus the expanded `Format` enum (`EXCEL`, `PREFIXMAP`, `PROTOBUF`, `SHACL`, `SQLDDL`, `SQLSCHEMA`, …) and `META_ONLY`. So that part is already in `main`.
- **The poetry.lock / `[tool.poetry]` changes** here are obsolete — `main` has since moved to a PEP 621 `[project]` layout with no poetry lockfile.
- **The two genuinely-missing pieces** — the GitHub URL fix and the `unittest`→`pytest` conversion — are carried forward in #257.

On the URL change specifically: I verified that `GITHUB_PATH_FOR` (branch/latest) and `GITHUB_IO_PATH_FOR` on `main` currently return **404** URLs because they omit the `linkml_model/` package segment; #257 fixes both (confirmed resolving with `curl`) and folds in @sneakers-the-rat's test-structure feedback (per-function tests, opt-in flag for the rate-limited GitHub API test).

Thanks @sierra-moxon for the original dig into this — the URL bug you flagged was real. /cc @kevinschaper @matentzn (re: triage).

PR #202, opened 2024-07-22
fix: decouple 'curie' uri from 'uri' uri

Silvanoc:
The URIs assigned to the different types identify the types themselves, as a consequence two different types with the same URI would be expected to be synonyms (different names with the same meaning).

This is the case for the types CURIE and URIorCURIE:
1. CURIE had the URI 'xsd:anyURI' (the same as for URIs).
2. URIorCURIE had the URI 'xsd:string' (the same as for strings).

This patch assign both CURIE and URIorCURIE their own unique URIs to differentiate the different types from each other.

### DIFF
--- a/linkml_model/model/schema/types.yaml
+++ b/linkml_model/model/schema/types.yaml
@@ -123,7 +123,7 @@ types:
       If you are authoring schemas in LinkML YAML, the type is referenced with the lower case "date_or_datetime".
 
   uriorcurie:
-    uri: xsd:anyURI
+    uri: linkml:UriOrCurie
     base: URIorCURIE
     repr: str
     description: a URI or a CURIE
@@ -131,7 +131,7 @@ types:
       If you are authoring schemas in LinkML YAML, the type is referenced with the lower case "uriorcurie".
 
   curie:
-    uri: xsd:string
+    uri: linkml:Curie
     base: Curie
     repr: str
     description: a compact URI


matentzn:
As a concerned citizen I would like to suggest this PR to be reviewed by at least 5 LinkML experts before it is merged. :P Changes to the base types whatever they are make me a bit nervous (thinking of effects on validation, linkml-dumper classes etc, not saying there are consequences, just... putting my worry out there).

Silvanoc:
> As a concerned citizen I would like to suggest this PR to be reviewed by at least 5 LinkML experts before it is merged. :P Changes to the base types whatever they are make me a bit nervous (thinking of effects on validation, linkml-dumper classes etc, not saying there are consequences, just... putting my worry out there).

@matentzn you are fully right and I share your concern, therefore this PR is in `Draft` status 😉

IMO this is a meaningful thing to do, because the current status is wrong. The URIs of the LinkML elements are their identifiers and having the same URI on multiple different elements means that they are the same. But URI ≠ CURIE ≠ URIorCURIE, right?

It is clear that when tools have "integrated" the error into their implementation fixing them might be the hardest task and we should not underestimate it.

I am more than happy to discuss:
1. If it is erroneous as I see it.
2. What is the impact in the ecosystem.
3. How to address this issue:
   * Fixing it
   * Ignoring it and expecting following consumers to work-around this issue

matentzn:
I totally agree! I think fixing this is a good idea..

Silvanoc:
@matentzn I have just realized that [this comment](https://github.com/linkml/linkml/issues/2213#issuecomment-2243099745) from @turbomam is probably signaling one place where this change might break something else.
So, yes, I will not move this PR from `Draft` until we have the clarity that we are not breaking anything inadvertently. 😄

mahdanoura:
@Silvanoc , I was checking the LinkML code base, and saw another place which _might_ be affected by your change. Sometimes variables that are of type `curie` are not assigned to the `type` `curie`, rather directly to a `string`.  As an example, in the LinkML [`SchemaDefinition`](https://github.com/linkml/linkml-runtime/blob/main/linkml_runtime/linkml_model/meta.py) class, 

https://github.com/linkml/linkml-runtime/blob/09b1f0c6a8bd55b794068879a0c4db0315c6f822/linkml_runtime/linkml_model/meta.py#L519C1-L527C65

mahdanoura:
@Silvanoc , I was checking the LinkML code base, and saw another type of place which _might_ be affected by your change. Sometimes variables that are of type `CURIE` are not assigned to the `type` `CURIE`, rather directly to a `string`.  As an example, in the LinkML [`SchemaDefinition`](https://github.com/linkml/linkml-runtime/blob/main/linkml_runtime/linkml_model/meta.py) class, 

[class_class_curie:](https://github.com/linkml/linkml-runtime/blob/09b1f0c6a8bd55b794068879a0c4db0315c6f822/linkml_runtime/linkml_model/meta.py#L527)
 variable is defined as `str`, rather than `CURIE`.

Silvanoc:
> @Silvanoc , another type of place which _might_ be affected by your change is in cases where variables of type `CURIE` are not assigned to that `type`, rather directly to a `string`. As an example, in the LinkML [`SchemaDefinition`](https://github.com/linkml/linkml-runtime/blob/main/linkml_runtime/linkml_model/meta.py) class,
> 
> [class_class_curie](https://github.com/linkml/linkml-runtime/blob/09b1f0c6a8bd55b794068879a0c4db0315c6f822/linkml_runtime/linkml_model/meta.py#L527) variable is defined as `str`. Therefore, changing the `CURIE` type as you suggest would not have any effect on these `str` assignments. Something like this would have been expected: `class_name: ClassVar[CURIE]`
> 
> However, I'm not deep into LinkML's code base and there could have been another reason for this assignment.

Nice catch! I will have to take a deeper look at it.

In general IMO the introduction of CURIEs, as it was done, has lots of issues.

- Partially due to the [ambiguity between CURIEs and URIs under certain conditions](https://www.w3.org/TR/2010/NOTE-curie-20101216/#sec_2.2.).
- Partially due to the unconventional type "UriOrCurie". JSON-Schema can easily handle it with `anyOf`, but Pydantic has a problem with it. And being lax by making URI, CURIE and URIorCURIE a string, makes Pydantic happy.
- In general handling these types is really complex and problematic due to the two above mentioned issues.

PR #207, opened 2024-11-08
Add URI or CURIE range documentation

amc-corey-cox:
This is just a tiny patch that adds a couple of comments to document that a URI or CURIE has to be created as a class in order to be used as a range for a slot.


pkalita-lbl (approved):
This appears to be inline with https://github.com/linkml/linkml/issues/2391

PR #208, opened 2024-11-22
Metamodel documentation site should use latest Jinja templates

sujaypatil96:
* This PR seeks to update the currently deployed linkml metamodel web documentation site (GitHub Pages) to use the latest Jinja templates available for use from the linkml library ([docgen templates](https://github.com/linkml/linkml/tree/main/linkml/generators/docgen)).
* It also pins the latest version of the mermaid in the mkdocs configuration because the mermaid mkdocs plugin which is supposed to pick up on a "compatible" version of mermaid does not do a very good job.
  * See: https://mkdocs-mermaid2.readthedocs.io/en/latest/#configuration
* Update multiline YAML syntax for one of the slots (`extra_slots_expression`) to use folded style (`>-`) instead of literal style (`|`) of block scalar syntax for reasons mentioned here: https://stackoverflow.com/questions/3790454/how-do-i-break-a-string-in-yaml-over-multiple-lines

CC: @sneakers-the-rat


sneakers-the-rat:
Seems like we should do this for all the poetry-based repos so we don't have to keep doing manual lockfile updates like this: https://github.com/linkml/linkml/bloab/main/.github/workflows/poetry-update.yaml

or else just do a `pip install` rather than a `poetry install` to always get the latest versions?

pkalita-lbl:
@sneakers-the-rat I recently started using Dependabot's [version updates feature](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/about-dependabot-version-updates) (as opposed to its longer-standing _security_ updates feature) in a few repositories to make periodic lock file updates. It's a lot simpler to set up than the bespoke thing we did in the `linkml/linkml` repo, and after using it for a few weeks it seems like it can accomplish the same thing (depending on how you configure it, of course). Here's an example PR that it produces: https://github.com/geneontology/gocam-py/pull/32. 

I don't think there's any particular rush in _replacing_ what we did in `linkml/linkml` with the Dependabot-based setup. But if I were going to set up something new in another repo I'd probably go the Dependabot route.

sneakers-the-rat:
Sure sure ya whatevs. I would call my action more of a hack than anything else, happy to do whatever, that looks more reliable

PR #210, opened 2024-12-24
Add Linked Open Vocabularies recommended ontology metadata

vincentkelleher:
What makes ontologies awesome is their reusability. To be reused an ontology needs to be discoverable and easily understandable through documentation (FAIR principles mention those requirements).

At Gaia-X we are working on making our LinkML generated ontology more discoverable and better documented through ontology metadata. The [Linked Open Vocabularies project](https://lov.linkeddata.es/dataset/lov) gives [recommendations](https://lov.linkeddata.es/Recommendations_Vocabulary_Design.pdf) about how an ontology can be described through metadata by using well established vocabularies such as Dublin Core Terms, Creative Commons and OWL.

This pull request proposes a solution to integrate slots for this metadata in the schema definition to then use those slots in the LinkML generators.


linkml_model/model/schema/meta.yaml
```diff
@@ -497,14 +498,6 @@ slots:
     in_subset:
       - BasicSubset
 
-  publisher:
```
sierra-moxon, line 500: we should probably deprecate this slot rather than delete.

linkml_model/model/schema/meta.yaml
```diff
@@ -2541,6 +2536,145 @@ slots:
...
+    slot_uri: cc:licence
+    in_subset:
+      - OwlProfile
+    exact_mappings:
+      - cc:licence
+    see_also:
+      - https://creativecommons.org/ns
+      - https://lov.linkeddata.es/Recommendations_Vocabulary_Design.pdf
+
+  creator:
```
sierra-moxon, line 2598: should `creator` and `contributor` be multivalued?

linkml_model/model/schema/meta.yaml
cmungall: agreed

linkml_model/model/schema/meta.yaml
```diff
@@ -49,6 +49,7 @@ prefixes:
   qudt: http://qudt.org/schema/qudt/
   cdisc: http://rdf.cdisc.org/mms#
   SIO: http://semanticscience.org/resource/SIO_
+  cc: http://creativecommons.org/ns#
```
cmungall, line 52: Let's stick to established prefixes, e.g bioregistry

linkml_model/model/schema/meta.yaml
```diff
@@ -2541,6 +2536,145 @@ slots:
...
+    slot_uri: dc:publisher
+    in_subset:
+      - OwlProfile
+    exact_mappings:
+      - dc:publisher
+    see_also:
+      - https://www.dublincore.org/specifications/dublin-core/dcmi-terms/#http://purl.org/dc/terms/publisher
+      - https://lov.linkeddata.es/Recommendations_Vocabulary_Design.pdf
+
+  preferredNamespacePrefix:
```
cmungall, line 2649: best to stick to existing conventions rather than mix (snake case)

linkml_model/model/schema/meta.yaml
```diff
@@ -2541,6 +2536,145 @@ slots:
...
+    slot_uri: dc:creator
+    in_subset:
+      - OwlProfile
+    exact_mappings:
+      - dc:creator
+    see_also:
+      - https://www.dublincore.org/specifications/dublin-core/dcmi-terms/#http://purl.org/dc/terms/creator
+      - https://lov.linkeddata.es/Recommendations_Vocabulary_Design.pdf
+
+  contributor:
```
cmungall, line 2615: we already have `contributors`, which can be applied at any level (schema, element, ...)

cmungall (changes requested):
I suggest breaking this up into smaller PRs, and doing a careful alignment with what already exists - e.g. we already have `contributors`, so no need for `contributor`

linkml_model/model/schema/meta.yaml
vincentkelleher: I had the exact same question while defining this but seemingly it's a single person according to Dublin Core:

image

I think it's possible to define multiple creators within an organization and reference it here 😊

linkml_model/model/schema/meta.yaml
vincentkelleher: I added this to be compliant with the [Linked Open Vocabularies](https://lov.linkeddata.es/dataset/lov/) standards but I imagine I can just change the behavior in the OWL generator to convert `contributors` to `contributor` ?

Then I'll have to find a way to manage multiple entries 🤔

linkml_model/model/schema/meta.yaml
vincentkelleher: Java/Typescript developer habits kicking in 🤭

linkml_model/model/schema/meta.yaml
vincentkelleher: It isn't deleted, you can find it at [line 2632](https://github.com/linkml/linkml-model/pull/210/files#diff-5fc807b5f16aafb921180aea704aa97a4c5a869eeff86344975963ba6b0bb49dR2632) 😇 

I didn't delete any existing attribute, I only modified them to add some information.

linkml_model/model/schema/meta.yaml
vincentkelleher: I can't understand which part of this prefix is troublesome 😌

Is it the use of `cc` which is too short ?
Or is it that Creative Commons cannot be considered as an established prefix ?

vincentkelleher:
> I suggest breaking this up into smaller PRs, and doing a careful alignment with what already exists - e.g. we already have `contributors`, so no need for `contributor`

What should be the granularity of those PRs ? One per attribute ?
I will have a closer look to see if I made some other alignment mistakes 👍

linkml_model/model/schema/meta.yaml
matentzn: The prefix is standard: https://bioregistry.io/registry/cc

The expansion as far as I can see should be https?

linkml_model/model/schema/meta.yaml
matentzn: I think they should both be multivalued. conceptually that does not translate to one property assertion with multiple values, but multiple property assertions, like

```
?schema dc:creator orcid:123 ;
   dc:creator orcid:124 .
```

That said, I think this is redundant with `contributors` see comment below.

linkml_model/model/schema/meta.yaml
matentzn: >  can just change the behavior in the OWL generator to convert contributors to contributor ?

Isnt this as easy as asserting slot_uri: dc:contributor? Need to check with @sierra-moxon to be sure.

linkml_model/prefixmap/meta.yaml
```diff
@@ -4,7 +4,9 @@
    "OIO": "http://www.geneontology.org/formats/oboInOwl#",
    "SIO": "http://semanticscience.org/resource/SIO_",
    "bibo": "http://purl.org/ontology/bibo/",
+   "cc": "http://creativecommons.org/ns#",
```
matentzn, line 7: https?

linkml_model/model/schema/meta.yaml
```diff
@@ -2541,6 +2536,145 @@ slots:
...
+      - https://www.dublincore.org/specifications/dublin-core/dcmi-terms/#http://purl.org/dc/terms/contributor
+      - https://lov.linkeddata.es/Recommendations_Vocabulary_Design.pdf
+
+  publisher:
+    description: >-
+      Publisher(s) of the vocabulary.
+      Recommended practice is to identify the contributor with a URI. 
+      If this is not possible or feasible, a literal value that identifies the publisher may be provided.
+    domain: schema_definition
+    range: string
```
matentzn, line 2638: not so keen on this, but I guess I will lose this battle.. (why not uriorcurie)?

linkml_model/model/schema/meta.yaml
```diff
@@ -2541,6 +2536,145 @@ slots:
...
+      - https://www.dublincore.org/specifications/dublin-core/dcmi-terms/#http://purl.org/dc/terms/rights
+      - https://lov.linkeddata.es/Recommendations_Vocabulary_Design.pdf
+
+  licence:
+    description: >-
+      Licence of the vocabulary. 
+      Recommended practice is to refer to a rights statement with a URI. 
+      If this is not possible or feasible, a literal value (name, label, or short text) may be provided.
+    domain: schema_definition
+    range: string
```
matentzn, line 2587: again, it would really help if we could get people that go all the way to developing a LinkML schema to use URIs for licenses, contributors and creators etc.

matentzn (commented):
I didnt look into redundancy but I think the proposed properties are useful. I have a little issue with anarchy in the ranges of the proposed properties, but I guess everyone can decided by themselves of they want to use stronger URIs (@sierra-moxon any idea what the rdf representation of a string typed literal that looks like a URI will look like? probably a string, right?)

PR #213, opened 2025-01-31
add simple_dict annotations to extensions

cmungall:
## Summary

Adds explicit `simple_dict_value: true` annotation to the `extension_value` slot in `extensions.yaml`, aligning the source schema with what's already present in derived artifacts (`meta.jsonld`, `meta.owl.ttl`).

## Motivation

The `extension` class has three slots: `extension_tag` (key), `extension_value` (required), and `extensions` (recursive self-reference, not required). Generators need to know which non-key slot holds the dict value to serialize extensions as simple `{tag: value}` dicts.

Currently this works via heuristic inference in `get_range_associated_slots()` — `extension_value` is the only non-key required slot, so it gets picked automatically. This annotation makes that intent explicit rather than relying on the fallback heuristic.

## Notes

- No functional change — generators already handle this correctly via inference
- The derived artifacts (`meta.jsonld`, `meta.owl.ttl`) already contain this annotation from a prior manual edit; this brings the source YAML into sync

EDIT: @matentzn added this PR description in the hope that it reflects @cmungall original intend.

### DIFF
--- a/linkml_model/model/schema/extensions.yaml
+++ b/linkml_model/model/schema/extensions.yaml
@@ -38,6 +38,8 @@ slots:
     alias: value
     required: true
     range: AnyValue
+    annotations:
+      simple_dict_value: true
 
 
 classes:


matentzn (changes requested):
@cmungall please add some basic text to the PR to understand the motivation and possible ramifications of this change.

matentzn (approved):
This adds `simple_dict_value: true` to `extension_value` in `extensions.yaml`. The `extension` class has three slots: `extension_tag` (key), `extension_value` (required), and `extensions` (recursive, not required). The thing is, `get_range_associated_slots()` in `helpers.py` already correctly identifies `extension_value` as the simple dict value slot via fallback heuristic #3: it's the only non-key required slot. So generators already handle this correctly without the annotation.

The `simple_dict_value` annotation in the derived artifacts (`meta.jsonld`, `meta.owl.ttl`) was likely added manually at some point — there's no code in linkml that auto-generates it during schema compilation. It's a stale/orphaned artifact, not evidence that the source YAML needs updating.

So this PR is purely defensive: making implicit behavior explicit. That's fine, but it's not fixing anything broken. The real question is whether we want to treat `simple_dict_value` annotations as required documentation or whether the heuristic-based inference is good enough. If we're going to annotate, we should probably do it consistently across all simple dict patterns, not just extensions.

amc-corey-cox (approved):
This appears to me to be a simple low-risk change. I'm inclined to approve.

But it does beg the question: should we make this change everywhere that could use this as well? What makes extensions special in this case? As things stand this slot can be in-lined as a simple dict without this change based on how linkml allows simple dict in-lining and this PR makes no functional change.

amc-corey-cox:
@cmungall Can you give a response to my and @matentzn reviews before we decide to merge this? I see nothing blocking this but I'm not sure if we actually need or want it.

PR #221, opened 2025-03-27
Add w3curie type and edit def of curie to be less restrictive

ialarmedalien:
Adding in the new w3curie type.

I tried to update the repo to use py3.12 with the current versions of linkml, linkml-runtime, and pyyaml so that I could regenerate the project files. Run `make gen-project` gave the following output:

```sh
# for all the files in the schema folder, run the gen-python command and output the result to the top
# level of the project.  In other repos, we'd include mergeimports=True, but we don't do that with
# linkml-model.
ValueError:  Unknown argument: slot_usage = {'dimensions': {'equals_number': 1}}
ValueError:  type "any_number" must declare a type base or parent (typeof)
cp staging/*.py linkml_model
poetry run gen-project -d staging --config-file gen_project_config.yaml linkml_model/model/schema/meta.yaml
Traceback (most recent call last):
[...snip a load of callback lines...]
ValueError: The ifabsent value `default_range` of the `range` slot could not be processed
make: *** [gen-project] Error 1
```

Looks like the various project generators, etc., have become more restrictive since `linkml-model` was last updated.

### DIFF
--- a/linkml_model/model/schema/types.yaml
+++ b/linkml_model/model/schema/types.yaml
@@ -31,7 +31,7 @@ types:
     base: str
     description: A character string
     notes: >-
-      In RDF serializations, a slot with range of string is treated as a literal or type xsd:string.  
+      In RDF serializations, a slot with range of string is treated as a literal or type xsd:string.
       If you are authoring schemas in LinkML YAML, the type is referenced with the lower case "string".
     exact_mappings:
       - schema:Text
@@ -132,14 +132,24 @@ types:
 
   curie:
     uri: xsd:string
+    base: URIorCURIE
+    repr: str
+    description: a prefixed ID, which may or may not conform to the w3 CURIE spec.
+    comments:
+      - Allows a wider range of characters than the stricter w3curie type.
+    notes: >-
+      If you are authoring schemas in LinkML YAML, the type is referenced with the lower case "curie".
+
+  w3curie:
+    uri: xsd:string
     base: Curie
     repr: str
-    description: a compact URI
+    description: A compact URI with a restricted character set; conforms to the w3 CURIE spec.
     comments:
       - in RDF serializations this MUST be expanded to a URI
       - in non-RDF serializations MAY be serialized as the compact representation
     notes: >-
-      If you are authoring schemas in LinkML YAML, the type is referenced with the lower case "curie".
+      If you are authoring schemas in LinkML YAML, the type is referenced with the lower case "w3curie".
     conforms_to: https://www.w3.org/TR/curie/
 
   uri:


Silvanoc (approved):

Silvanoc:
Would it be possible to have something like w3safecurie too as a subtype of curie? That way we could have something like CurieOrSafeCurie too.

ialarmedalien:
I don't have merge permissions -- @cmungall @sierra-moxon can you approve/merge?

Silvanoc:
> I don't have merge permissions -- @cmungall @sierra-moxon can you approve/merge?

@ialarmedalien I can merge, but the (undocumented) policy says that only the core-team is allowed to merge. And I'm not part of it.

I started a [discussion to get this kind of policies, workflows and processes documented](https://github.com/orgs/linkml/discussions/2518). But up to now I didn't get any resonance in the community 🤷

nlharris:
@Silvanoc good point about "thumbs up" vs. "votes". I think for a thumbs-up to count as a vote, it has to be on the issue description (i.e., the first "comment", as opposed to any subsequent comments) - is that correct?

Regarding this PR, I realize you are eager to get it merged. However, we are extra-careful about changes that affect the metamodel. This PR has no automated tests to make sure it doesn't break anything else, and it's also not clear whether anyone manually tested how this change affects things when imported into an existing LinkML model repository (such as Biolink) or any downstream repo in the LinkML ecosystem. To avoid potential problems caused by merging this PR, we need people who are deeply familiar with the metamodel and the whole codebase to review it. (This would of course be easier if there were tests included.)

I hear you on the frustration about lack of clarity in the process for reviewing and merging PRs. This is something that many open-source projects run into as they grow. Thank you for opening Discussion #2518 and including examples of some other open-source projects' contribution protocols. We will definitely look into those.

ialarmedalien:
@nlharris Not sure what kind of tests should be added -- the PR registers a new data type, derived from existing data types, to the data model; it does not add any functions or methods, which would require tests. There are no extant tests for this part of the model that new tests could be modelled on. I'm happy to add some if you want to suggest what they should do.

ialarmedalien:
> OSS projects typically start opening the source code, but are not open communities. At that point there might not be a real need to processes, as long as the community knows some otherwise established or common processes. But OSS projects eventually become also open communities. And at that point openly documented processes become important for a smooth community interaction. IMO LinkML has reached that point for a long time now and should really document at least the basics.

Agree completely. As @Silvanoc alluded to in an earlier comment, there seems to be a "core" LinkML developer team, but it's not clear who is part of that team beyond Chris M and Sierra. Having established and documented procedures for PR triage would also be beneficial for maintaining positive sentiment in the community.

PR #224, opened 2025-05-20
Minimal changes to support windows

dalito:


dalito:
I cannot re-run the failed job here. It failed due to github-api rate limit. 
...either I lack permissions in this repo or re-starting runs requires `workflow_dispatch` which is only added in this PR.

matentzn:
@sujaypatil96 please review this PR asap - its been open for a long while!

@dalito Thanks for the patience!

matentzn:
@dalito thanks I will change the permissions as soon as I have them! I pinged @sujaypatil96 again to review, so should be soon.

PR #225, opened 2025-05-20
Update GitHub Actions workflows (sync with other repos in org)

sujaypatil96:
Fixes https://github.com/linkml/linkml/issues/2720

Note: Updated the docs building/publishing workflow to use Python 3.10 to build the docs instead of Python 3.9.

### DIFF
--- a/.github/workflows/build-publish-documentation.yaml
+++ b/.github/workflows/build-publish-documentation.yaml
@@ -69,9 +69,11 @@ jobs:
           # generate HTML from markdown and put into dev version
           poetry run mike deploy dev
           
+          # ensure the main branch is available locally
+          git fetch origin main
           # switch to gh-pages branch, copy over the latest 404.html from main
           git checkout --force gh-pages
-          git checkout --force main -- 404.html
+          git checkout --force origin/main -- 404.html
           git commit -m "Add 404.html to root" || echo "No changes to 404.html to commit"
           
           # push changes to gh-page branch
--- a/.github/workflows/build-publish-documentation.yaml
+++ b/.github/workflows/build-publish-documentation.yaml
@@ -29,13 +29,6 @@ jobs:
         with:
           fetch-depth: 0 # fetch all commits/branches so that mike works
 
-      - name: Setup Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v5.6.0
-        id: setup-python
-        with:
-          python-version: ${{ matrix.python-version }}
-          cache: 'poetry'
-              
       #----------------------------------------------
       #          install & configure poetry
       #----------------------------------------------
@@ -43,6 +36,13 @@ jobs:
         run: |
           pipx install poetry
           pipx inject poetry poetry-dynamic-versioning
+
+      - name: Setup Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v5.6.0
+        id: setup-python
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache: 'poetry'
 
       #----------------------------------------------
       #             install dependencies
--- a/.github/workflows/build-publish-documentation.yaml
+++ b/.github/workflows/build-publish-documentation.yaml
@@ -13,10 +13,6 @@ jobs:
   github-pages:
     runs-on: ubuntu-latest
 
-    strategy:
-      matrix:
-        python-version: [ "3.10" ]
-
     permissions:
       contents: write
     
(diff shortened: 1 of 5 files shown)


.github/workflows/build-publish-documentation.yaml
```diff
@@ -18,40 +25,30 @@ jobs:
...
       #          install & configure poetry
       #----------------------------------------------
       - name: Install Poetry
-        uses: snok/install-poetry@v1.3
-        with:
-          virtualenvs-create: true
-          virtualenvs-in-project: true
+        run: |
+          pipx install poetry
+          pipx inject poetry poetry-dynamic-versioning
```
dalito, line 38: The repo is not yet setup for dynamic-versioning so you don't need to install it.

.github/workflows/main.yaml
```diff
@@ -6,59 +6,48 @@ name: Build and test linkml-model
...
       #          install & configure poetry
       #----------------------------------------------
       - name: Install Poetry
-        uses: snok/install-poetry@v1.3
-        with:
-          virtualenvs-create: true
-          virtualenvs-in-project: true
+        run: |
+          pipx install poetry
+          pipx inject poetry poetry-dynamic-versioning
```
dalito, line 35: The repo is not yet setup for dynamic-versioning so you don't need to install it.

.github/workflows/release.yaml
```diff
@@ -14,20 +14,19 @@ jobs:
...
+          python-version: 3.10
 
       - name: Install Poetry
-        uses: snok/install-poetry@v1.3
-        with:
-          virtualenvs-create: true
-          virtualenvs-in-project: true
+        run: |
+          pipx install poetry
+          pipx inject poetry poetry-dynamic-versioning
```
dalito, line 29: Again. The repo is not yet setup for dynamic-versioning so you don't need to install it.

.github/workflows/build-publish-documentation.yaml
```diff
@@ -72,9 +69,11 @@ jobs:
           # generate HTML from markdown and put into dev version
           poetry run mike deploy dev
           
+          # ensure the main branch is available locally
+          git fetch origin main
           # switch to gh-pages branch, copy over the latest 404.html from main
           git checkout --force gh-pages
-          git checkout --force main -- 404.html
+          git checkout --force origin/main -- 404.html
```
dalito, line 76: I use mike for multi-versioning in [nfdi4cat/pid4cat-model](https://github.com/nfdi4cat/pid4cat-model) and the checkout acrobatics and the copying of the 404.html here are strange. I added a [proper 404.html](https://github.com/nfdi4cat/pid4cat-model/blob/gh-pages/404.html) to the gh-pages branch and never hat to touch it again.

However, if it works here then maybe leave it as it is.

.github/workflows/build-publish-documentation.yaml
```diff
@@ -5,11 +5,18 @@ on:
...
+    branches:
+      - main
 
 jobs:
   github-pages:
     runs-on: ubuntu-latest
 
+    strategy:
+      matrix:
+        python-version: [ "3.10" ]
```
pkalita-lbl, line 18: Is there any reason to use a `matrix` here when we only will ever do this using one Python version?

.github/workflows/build-publish-documentation.yaml
```diff
@@ -18,40 +25,30 @@ jobs:
...
+          cache: 'poetry'
 
       #----------------------------------------------
-      # install dependencies if cache does not exist 
+      #             install dependencies
       #----------------------------------------------
       - name: Install dependencies
-        if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
-        run: poetry install --no-interaction --no-root
+        run: poetry install --no-interaction --with docs --no-root --all-extras
```
pkalita-lbl, line 51: As far as I can tell `pyproject.toml` for this project doesn't specify any extras and the `docs` group is not marked as optional. So the `--with docs` and `--all-extras` flags aren't doing anything here.

.github/workflows/build-publish-documentation.yaml
pkalita-lbl: I'm responsible for the checkout acrobatics here! 😂 

I'm afraid I don't quite remember all the details on why it is the way it is. There is a brief comment in https://github.com/linkml/linkml-model/pull/168 about it. But ultimately I think I just didn't like the idea of a file _only_ living in the `gh-pages` branch where it could be potentially forgotten about.

.github/workflows/release.yaml
sierra-moxon: I'm not sure if this was a particular design decision to be honest.  It could be that linkml-model repo was created before we started using dynamic versioning and we didn't bring it back up to "standard." 

We can do dynamic versioning in another PR.

.github/workflows/build-publish-documentation.yaml
sujaypatil96: Good point. I saw this pattern in workflows in other repos like linkml – https://github.com/linkml/linkml/blob/main/.github/workflows/doc-pages.yaml#L10-L11 and I thought that we should change it, but left it this way to be consistent with other repos.

But yes you're absolutely right, we're going to be building and publishing the docs using only one version of Python and there's no reason to have a python-version matrix like we have above for that.

I'll go ahead and make the change to correct that. Will make PR's to fix this in other repos as well.

.github/workflows/build-publish-documentation.yaml
sujaypatil96: Awesome! thanks for picking up on that Patrick. Will correct it.

pkalita-lbl (approved):

.github/workflows/release.yaml
dalito: I would also suggest to add it in another PR.

PR #241, opened 2026-02-26
Update metamodel files

matentzn:


sneakers-the-rat (approved):
gotta love the +10billion/-10billion everytim from the unsorted rdf, one day the good ttl canonicalizer will make it into rdflib. 

thanks for handling this!!!

PR #246, opened 2026-04-15
Fix invalid slot_usage on axis slot in array.yaml

### LINKED ISSUE #3389: Fix invalid `slot_usage` on slot definition in `array.yaml`
`slot_usage` is used on the `axis` slot definition in `array.yaml`, but `slot_usage` is a class-level construct — it's not valid on slots. This causes `gen-python --genmeta` to fail with `TypeError: SlotDefinition.__init__() got an unexpected keyword argument 'slot_usage'`.

## Context

In `linkml_model/model/schema/array.yaml`, lines 107–111:

```yaml
  axis:
    range: NDArray
    slot_usage:
      dimensions:
        equals_number: 1
```

`slot_usage` refines inherited slots within a class. Putting it on a slot definition is structurally invalid. The Makefile's `gen-py` target already excludes `extended_types.yaml` for a similar reason ("schema is not ready for Python code generation yet"), but `array.yaml` is not excluded and breaks the build.

## Expected behaviour

`array.yaml` should be valid LinkML and pass `gen-python --genmeta` without errors.

## Suggested approach

Remove `slot_usage` from the `axis` slot definition. The intent — constraining `dimensions` to 1 when used as an axis — should be expressed at the class level instead, e.g. by adding `slot_usage` to `DataArray` (which uses the `axis` slot), or via a class rule.

Alternatively, if `array.yaml` is genuinely not ready for code generation (it is marked `status: testing`), exclude it in the Makefile alongside `extended_types.yaml` and document why.

matentzn:
You cant have a slot_usage on a SLOT

Fixes https://github.com/linkml/linkml/issues/3389

### DIFF
--- a/linkml_model/model/schema/array.yaml
+++ b/linkml_model/model/schema/array.yaml
@@ -55,6 +55,14 @@ classes:
           an ordered sequence of values. The elements also have an *array interpretation*, where each
           element has a unique index which is determined by array_linearization_order
 
+  OneDimensionalSeries:
+    description: >-
+      an NDArray whose dimensionality is constrained to 1
+    is_a: NDArray
+    slot_usage:
+      dimensions:
+        equals_number: 1
+
   DataArray:
     description: >-
       a data structure containing an NDArray and a set of one-dimensional series that are used to label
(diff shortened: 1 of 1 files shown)


rly:
This looks fine to me. Tagging @sneakers-the-rat for a second look as they are also well-versed in the array model.

matentzn:
Thanks @rly !

sneakers-the-rat:
seems fine to me :)

sneakers-the-rat (approved):

sneakers-the-rat:
Screenshot 2026-04-16 at 6 10 16 PM

rly (approved):

PR #249, opened 2026-04-27
Fix mistyping in MatchQuery description

plokhotnyuk:


tests/input/meta.yaml
matentzn, line 1: This is the only change that matters.

matentzn (dismissed):

matentzn (approved):
Thank you @plokhotnyuk !

plokhotnyuk:
@matentzn Thanks for cleaning up my PR!

PR #250, opened 2026-04-28
Rebuild 1.11.0 rc2 model version

matentzn:


amc-corey-cox (approved):
This has the changes we need for https://github.com/linkml/linkml/pull/3388 and doesn't have anything else I wouldn't expect. Looks good to me.

kevinschaper (approved):
👍

PR #253, opened 2026-05-04
Refresh model files for 1.11.0-rc3

matentzn:
Regenerates the generated artifacts under `linkml_model/` against the current source schemas in `linkml_model/model/schema/` for the upcoming 1.11.0-rc3 release. Two real changes plus some cleanup:

1. **Picks up the `permissivle` → `permissible` typo fix from #249.** That PR fixed the typo in `meta.yaml` but didn't refresh the downstream generated artifacts. This PR propagates the fix into `meta.py`, `jsonld/meta.jsonld`, `jsonschema/meta.schema.json`, `protobuf/meta.proto`, `owl/meta.owl.ttl`, `shacl/meta.shacl.ttl`, and `sqlschema/meta.sql`.

2. **Populates `linkml_model/array.py`.** The file existed but was empty — `array.yaml` has been part of the schema source for a while but the Python module was never generated from it (see #251). Now it ships properly (285 lines).

3. **Revives stale `linkml_model/rdf/` and `linkml_model/sqlddl/`.**  These were removed from the gen-project defaults at some time back in 2023 - I hereby add them back.


linkml_model/rdf/README.md
```diff
@@ -1,3 +0,0 @@
-# rdf
```
amc-corey-cox, line 1: Removing `rdf/` and `sqlddl/` leaves dangling references in `linkml_model/linkml_files.py`, which is part of the public API:

- `Format.NATIVE_RDF`, `Format.RDF`, `Format.SQLDDL` (enum entries, lines 41/48/52)
- `_Path.NATIVE_RDF`, `_Path.RDF`, `_Path.SQLDDL` (`_PathInfo` mappings, lines 67/74/78)
- `Format.SQLDDL` is in the `META_ONLY` tuple (line 100)
- `ModelFile.model_rdf()` and `ModelFile.rdf()` accessors (lines 230/250)

After this merges, calls like `LOCAL_PATH_FOR(Source.META, Format.RDF)` will silently return paths to files that no longer exist. `tests/test_linkml_files.py:52` still references `Format.RDF`; the test happens to pass only because it asserts on URL-string construction, not file existence.

Could you either drop these enum entries / accessors (with a short deprecation note in CHANGELOG/release notes) or keep the directories? Whichever way, the in-repo API surface should match what's on disk.

amc-corey-cox (approved):
This looks good to me. Claude pointed out to me that we now have a few files named in linkml_files.py that we just deleted so they now don't exist. Might be nice to clean that up but it's just a nit. This is ready to go.

linkml_model/rdf/README.md
matentzn: I will wait for @dalito to re-assure me that removal is the right path over "revival" and then I will deal with this! Thank you @amc-corey-cox

linkml_model/array.py
```diff
@@ -0,0 +1,284 @@
...
+    extended_str
+)
+from rdflib import (
+    Namespace,
+    URIRef
+)
+
+from .types import Integer, String
+
+metamodel_version = "1.7.0"
```
rly, line 58: This and the other generated Python files that I looked at report a metamodel version of 1.7.0. I think this should be updated to 1.11.0.

rly:
One minor comment above about `metamodel_version` in the python files, but otherwise the array parts that I am knowledgeable about look good to me.

linkml_model/rdf/README.md
matentzn: I decided to change strategy and revive rather than bury these. gen-project had dropped them from the defaults a while back, and so I added them back to the build now.

matentzn:
> One minor comment above about metamodel_version in the python files, but otherwise the array parts that I am knowledgeable about look good to me.

I did the change now but it unfortunately wont permeate all the way through until the NEXT release. This is because some of the generated artefacts use the metamodel version vendored into the meta.py INSIDE the runtime. Once this is merged and updated in the runtime, this loop will close a bit more, but it is an annoying side effect that the "metamodel version" used to generate the metamodel is NOT the same as the meta model version being updated.

linkml_model/array.py
matentzn: I punt this to here: https://github.com/linkml/linkml/issues/3479

rly: Thanks!

linkml_model/array.py
```diff
@@ -0,0 +1,284 @@
...
+
+        super().__post_init__(**kwargs)
+
+
+# Enumerations
+class ArrayLinearizationOrderOptions(EnumDefinitionImpl):
+    """
+    Determines how a linear contiguous representation of the elements of an array map to array indices
+    """
+    COLUMN_MAJOR_ARRAY_ORDER = PermissibleValue(
```
rly, line 223: Sorry, I'm looking at this in more detail now. Claude flagged this -- `PermissibleValue` is not imported or defined in this file. Unless I am missing something, I don't think this file can be imported. Is that OK? The top-level comment says "model is not intended to be imported directly". I don't fully understand how this file is used, and I suppose since this is the first time it is generated, it hasn't been used at all!

linkml_model/array.py
```diff
@@ -0,0 +1,284 @@
...
+        text="COLUMN_MAJOR_ARRAY_ORDER",
+        description="""An array layout option in which the elements in each column is stored in consecutive positions, or any generalization thereof to dimensionality greater than 2""",
+        meaning=GOM["columnMajorArray"])
+    ROW_MAJOR_ARRAY_ORDER = PermissibleValue(
+        text="ROW_MAJOR_ARRAY_ORDER",
+        description="""An array layout option in which the elements in each row is stored in consecutive positions, or any generalization thereof to dimensionality greater than 2""",
+        meaning=GOM["rowMajorArray"])
+
+    _defn = EnumDefinition(
```
rly, line 232: Same for `EnumDefinition`.

linkml_model/array.py
```diff
@@ -0,0 +1,284 @@
...
+    a data structure consisting of a collection of *elements*, each identified by at least one array index tuple.
+    """
+    _inherited_slots: ClassVar[list[str]] = []
+
+    class_class_uri: ClassVar[URIRef] = LINKML["NDArray"]
+    class_class_curie: ClassVar[str] = "linkml:NDArray"
+    class_name: ClassVar[str] = "NDArray"
+    class_model_uri: ClassVar[URIRef] = LINKML.NDArray
+
+    elements: Union[Union[dict, Any], list[Union[dict, Any]]] = None
```
rly, line 97: The type hint says that `elements` is required but there is no
```python
if self._is_empty(self.elements):
  self.MissingRequiredField("elements")
```
like in `DataArray.__post_init__`. Probably an issue with `pythongen.py` but just flagging it as something that appears to be missing. Not blocking if this file isn't used anyway

amc-corey-cox (approved):
This looks good to me now.

matentzn:
@rly I filed a follow up issue for all your observations and assigned @amc-corey-cox : https://github.com/linkml/linkml/issues/3483

These were broken before - PR just surfaces the brokenness - and we need to RC release out now! THANKS 1 million for the reviews!

PR #254, opened 2026-05-05
Add missing prefix definitions

plokhotnyuk:


matentzn (changes requested):
I approve but will block this PR until we have 1.11 out the door! Thank you!

Ostrzyciel:
> I approve but will block this PR until we have 1.11 out the door! Thank you!

@matentzn can we merge this now?

matentzn (approved):

matentzn (approved):
This is great, thanks. Will merge. Note:

datasets.yaml:284 has slot_uri: prov:wasDerivedFrom, but prov is neither declared nor added here; might be worth a follow up. I will get a second reviewer but will make sure its merged asap.

dalito (approved):
Looks good to me. 

Interesting that it was not fixed long ago. Probably because these are prefixes that rdflib loads by default(?). I also spotted that the schemas have prefixes under `emit_prefixes:` that they do not declare e.g. https://github.com/plokhotnyuk/linkml-model/blob/df58102f1f53c847843964cff3a052666f91b9bd/linkml_model/model/schema/datasets.yaml#L37

PR #259, opened 2026-06-24
Resolve ambiguity on how to handle attributes

### LINKED ISSUE #3669: gen-jsonld - class attribute gets duplicated into attribute and class-specific-slot
**Describe the bug**

The JSON-LD document generated of a LinkML YAML document contains somehow duplicated information:
* a class-specific-slot
* an inline attribute

~No 100% sure which one is the desired one, but it should be only one of both.~

~According [this comment in the Meta-Model `ClassDefinition`](https://linkml.io/linkml-model/latest/docs/attributes/#comments):~

~> attributes are an alternative way of defining new slots. An attribute adds a slot to the global space in the form __ (lower case, double underscores). Attributes can be specialized via slot_usage.~

~What IMO is equivalent to saying "`attributes` are only syntactic sugar for class-specific slots". Perhaps the specification should also state is clearer in the description of the slot `attributes`'.~

Parent issue #3670 created to get first clarity on the specification to know what is the expected output.

**Version of LinkML you are using** 1.11.1

**Please provide a schema (and if applicable, a data file) that replicates the issue**
LinkML YAML schema:
```yaml
id: http://example.com/schema_a
name: schema_a
prefixes:
  schema_a: http://example.com/schema_a/
  linkml: https://w3id.org/linkml/
default_prefix: schema_a
imports:
  - linkml:types

classes:
  Person:
    attributes:
      age:
        range: float
```

Subset of resulting JSON-LD schema:
```json
{
  <...>
  "slots": [
    {
      "name": "person__age",
      "from_schema": "http://example.com/schema_a",
      "slot_uri": "http://example.com/schema_a/age",
      "alias": "age",
      "owner": "Person",
      "domain_of": [
        "Person"
      ],
      "range": "float",
      "@type": "SlotDefinition"
    }
  ],
  "classes": [
    {
      "name": "Person",
      "definition_uri": "http://example.com/schema_a/Person",
      "from_schema": "http://example.com/schema_a",
      "slots": [
        "person__age"
      ],
      "slot_usage": {},
      "attributes": [
        {
          "name": "age",
          "range": "float",
          "@type": "SlotDefinition"
        }
      ],
      "class_uri": "http://example.com/schema_a/Person",
      "@type": "ClassDefinition"
    }
  ],
  <...>
}
```

Please notice that the _Person_'s _attribute_  `age` has been "promoted" to the class-specific-slot `person__age`. And notice also that `age` has been also kept as the inline attribute `age`.

Silvanoc:
This patch fixes the `attributes` specification for more clarity on how to handle `attributes` converting to other formats.

Before this patch the `attributes` specification had a comment that gave the impression that `attributes` are only syntactic sugar for class-scoped slots (slots that are only available to be used by a class and its descendants) and specifying the conversion of `attributes` into `slots` renaming them to `<class-name>__<attribute-name>`.

This patch modifies that comment so that `attributes` are kept as 1st-class-citizens even on conversions.

Fixes: linkml/linkml#3669

### DIFF
--- a/linkml_model/model/schema/meta.yaml
+++ b/linkml_model/model/schema/meta.yaml
@@ -1231,11 +1231,11 @@ slots:
     multivalued: true
     range: slot_definition
     inlined: true
-    description: Inline definition of slots
+    description: Inline definition of class-scoped slots
     comments:
       - >-
-        attributes are an alternative way of defining new slots.  An attribute adds a slot to the global space in the
-        form <class_name>__<slot_name> (lower case, double underscores).  Attributes can be specialized via slot_usage.
+        Attributes are an alternative way of defining class-specific slots.
+        They are equivalent to slots that are only available within the scope of class and its descendants.
     in_subset:
       - SpecificationSubset
       - MinimalSubset


gouttegd (dismissed):
I am (obviously) not an authority on the meta-model, but FWIW, I agree with this.

The `<class_name>__<slot_name>` bit is nothing but an implementation detail. This is how some parts of LinkML-Py deal with attributes, but this is not required and there are other ways to deal with them.

linkml_model/model/schema/meta.yaml
```diff
@@ -1231,11 +1231,11 @@ slots:
...
     range: slot_definition
     inlined: true
-    description: Inline definition of slots
+    description: Inline definition of class-scoped slots
     comments:
       - >-
-        attributes are an alternative way of defining new slots.  An attribute adds a slot to the global space in the
-        form <class_name>__<slot_name> (lower case, double underscores).  Attributes can be specialized via slot_usage.
+        Attributes are an alternative way of defining class-specific slots.
+        They are equivalent to slots that are only available within the scope of class and its descendants.
```
matentzn:
```suggestion
        They are equivalent to slots that are only available within the scope of the class and its descendants.
```

linkml_model/model/schema/meta.yaml
```diff
@@ -1231,11 +1231,11 @@ slots:
     multivalued: true
     range: slot_definition
     inlined: true
-    description: Inline definition of slots
+    description: Inline definition of class-scoped slots
     comments:
       - >-
-        attributes are an alternative way of defining new slots.  An attribute adds a slot to the global space in the
-        form <class_name>__<slot_name> (lower case, double underscores).  Attributes can be specialized via slot_usage.
```
matentzn, line 1238: Is it necessary to remove this phrase "Attributes can be specialized via slot_usage."? Seems correct to me..

linkml_model/model/schema/meta.yaml
```diff
@@ -1231,11 +1231,11 @@ slots:
     multivalued: true
     range: slot_definition
     inlined: true
-    description: Inline definition of slots
+    description: Inline definition of class-scoped slots
```
matentzn, line 1234: Does it make sense to change that here as well? https://github.com/linkml/linkml-model/blob/35c91fb01382bb1c562e6179823241e836b29519/linkml_model/model/docs/specification/03schemas.md?plain=1#L174

linkml_model/model/schema/meta.yaml
matentzn: What would you expect `child_via_attr` to come out as here?

```yaml
slots:
  parent_slot: {abstract: true, required: true, pattern: "^x.*", multivalued: true}
  child_via_slot: {is_a: parent_slot}
classes:
  ViaSlots:
    slots: [child_via_slot]
  ViaAttrs:
    attributes:
      child_via_attr: {is_a: parent_slot}
```

Same parent, same `is_a`. `child_via_slot` comes out required, patterned and multivalued; `child_via_attr` comes out with none of them — and per the spec that's correct, not a bug. Still "equivalent"?

linkml_model/model/schema/meta.yaml
gouttegd: Isn’t that implied by the fact that attributes are “equivalent to slots” – since slots can be specialised via `slot_usage`?

Though on second thought, at least one generator falsely believes that `slot_usage` should _not_ be used to refine attributes (https://github.com/linkml/linkml/issues/3438), so maybe keeping that explicit statement would in fact be helpful…

Silvanoc: I don't remember if I removed it for a specific reason or it was just an erroneous "delete until end of line". I'll double check.

linkml_model/model/schema/meta.yaml
Silvanoc: I would expect `child_via_slot` and `child_via_attr` to have exactly the same attributes.

I don't know if this comes from the `descendants` part. If that's the case, it refers to the class declaring the attribute and its subclasses. It has nothing to do with a slot's `is_a` relationship.

linkml_model/model/schema/meta.yaml
Silvanoc: Perhaps a rewording like

> Attributes are slots and such can be also modified via `slot_usage`.

might be helpful here. Assuming that it's true, what I still have to double check.

gouttegd: > Assuming that it's true, what I still have to double check.

But the question is, _where_ do you check that? AFAIK, the only place in the meta-model where anything is said about `slot_usage` applying to attributes is here. The definition of `slot_usage` only says that it is “the refinement of a slot in the context of the containing class definition” – which _may_ be interpreted as being applicable to attributes given that an attribute is nothing more than a type of slot, but who knows?

The “reference implementation” (LinkML-Py) is not helpful here since its different components cannot agree with each other. `SchemaView` is of the position that `slot_usage` does apply to attributes. But the Python generator (which is not based on `SchemaView`) disagrees.

PR #261, opened 2026-07-02
feat(openenum): add is_open boolean slot to enum expressions

noelmcloughlin:
Wait for https://github.com/linkml/linkml-model/pull/265 and https://github.com/linkml/linkml-model/pull/266

This PR adds `is_open` boolean slot to enum expressions in the meta model.

See https://github.com/linkml/linkml/issues/127 and https://github.com/linkml/linkml/pull/3712


linkml_model/model/schema/meta.yaml
```diff
@@ -73,7 +73,7 @@ imports:
   - linkml:mappings
   - linkml:extensions
   - linkml:annotations
-#  - linkml:validation
+  #  - linkml:validation
   - linkml:units
```
matentzn, line 77: Can we remove any style changes unrelated to this PR and move them to a new one? I would like to be strict on meta-model changes for `git blame` etc.

linkml_model/model/schema/meta.yaml
```diff
@@ -650,6 +649,25 @@ slots:
     exact_mappings:
       - cdisc:PermissibleValue
 
+  is_open:
+    domain: enum_expression
```
matentzn, line 654: Maybe add `status: testing` to this slot, as per convention for newly introduced slots?

linkml_model/model/schema/meta.yaml
```diff
@@ -650,6 +649,25 @@ slots:
...
+    range: boolean
+    description: >-
+      Whether the enumeration is open, meaning that values outside of the set of
+      permissible values are allowed. If true, the enumeration is open and data values
+      that are not in the set of permissible values are permitted (though they SHOULD be
+      drawn from the permissible values if available). If false (the default), the enumeration
+      is closed and data values MUST be drawn from the set of permissible values.
+    comments:
```
matentzn, line 662: Worth adding a `see_also` to the NIST metaschema `allow-other` spec (https://pages.nist.gov/metaschema/specification/syntax/constraints/#allow-other) since the comments already reference OSCAL?

linkml_model/model/schema/meta.yaml
```diff
@@ -650,6 +649,25 @@ slots:
...
+      - An open enumeration is analogous to the `allow-other="yes"` construct in the OSCAL Metamodel.
+      - When an enum is open, the permissible values are advisory - they document the recommended
+        or expected values but do not constrain the range. In JSON Schema this maps to a plain string.
+      - When an enum is closed (the default) and has no permissible values (and no other means of
```
matentzn: the default? How?

Did we discuss whether we should add `ifabsent: false`?

linkml_model/model/schema/meta.yaml
```diff
@@ -650,6 +649,25 @@ slots:
     exact_mappings:
       - cdisc:PermissibleValue
 
+  is_open:
```
matentzn, line 653: Which issue are we using to track this slot .. and its implementation into runtime? I would kinda like to see the issues references in the schema as well (see_also).

linkml_model/model/schema/meta.yaml
```diff
@@ -650,6 +649,25 @@ slots:
...
+      is closed and data values MUST be drawn from the set of permissible values.
+    comments:
+      - An open enumeration is analogous to the `allow-other="yes"` construct in the OSCAL Metamodel.
+      - When an enum is open, the permissible values are advisory - they document the recommended
+        or expected values but do not constrain the range. In JSON Schema this maps to a plain string.
+      - When an enum is closed (the default) and has no permissible values (and no other means of
+        computing them), this is a schema error.
+    in_subset:
+      - SpecificationSubset
+      - BasicSubset
```
matentzn: Are we sure we want to add it to basicsubset already before we have actual support any where in runtime?

matentzn (dismissed):

linkml_model/model/schema/meta.yaml
noelmcloughlin: Added issue link

linkml_model/model/schema/meta.yaml
noelmcloughlin: good point - added `ifabsent`

linkml_model/model/schema/meta.yaml
noelmcloughlin: resolved

sneakers-the-rat:
The basic idea makes total sense to me. I just wonder if we want to have a single way of declaring this across the few places we want similar behavior - e.g. `extra_slots` allows for a boolean "yes others are allowed"/"no others are forbidden" as well as specifying an expression that other values must match. Do we want that here? We also need to add a similar thing at the SchemaDefinition level to specify openness of the whole schema/specify class/slot expressions to match extra members as well.

`extra_slots` as a name is too specific, I like `is_open` for a bool, the most generic name might just be `extra`.

Basically I am in favor of this but just wondering if we want to have a similar mechanism/pattern across all "openness" specs, or if there is a reason this one should be a boolean only

noelmcloughlin:
I am just now thinking maybe `is_open` is bad name - I recall other discussions where someone (Chris?) worried it might get conflated with "open world" .. if push back comes then `allow-other` has precedent in Oscal Metamodel.

But issue https://github.com/linkml/linkml/issues/127 proposes 'open  enums' via `is_open` meta slot.

I noted https://github.com/linkml/linkml/issues/1595 discusses `extensible` meta slot also - kinda related.

sneakers-the-rat:
This has been a general naming tension between how I normally think and how linkml does names - I generally tend towards more general names whose meaning varies by context, and linkml tends to have specific names and only use them in one or a few contexts in which is has a single meaning. So where I would probably gravitate towards a single `extra:` term that could be used for slots, classes, enum items, etc., I think the linkml way would have a term per type of extra thing, `extra_classes`, `extra_enum_items`, etc.

`is_open` is probably the name I would reach for too for a boolean name, but it sort of doesn't have an obvious corrolary for the "middle" switch - true is "allow all," false is "allow none," but what would you call "allow some" (I.e. expressed in some subschema). `open_to` maybe? This was a consideration I remember finding tricky when landing on `extra_*` and having sub-terms within it, which is not the most elegant solution. 

But anyway, need to decide if we want to have common mechanism/pattern for declaring "all/none/some" openness, or if sometimes binary and sometimes trinary is fine. That also narrows down which names work.

matentzn:
As this issue is slightly beyond my current ability to judge, @amc-corey-cox has volunteered to thinking through the implications and presenting a short version of this at the next developer call in one week. Sorry for the delay - please keep pushing if we slow you down; I can find more aggressive alternative review processes. 🚀

linkml_model/model/schema/meta.yaml
```diff
@@ -650,6 +650,29 @@ slots:
     exact_mappings:
       - cdisc:PermissibleValue
 
+  is_open:
```
amc-corey-cox, line 653: A boolean feels right here. `id_prefixes_are_closed` (line 249) is the precedent: one slot builds the set, another says if it's exhaustive. `is_open: true` is also the top case of a future `open_to: <expr>`, so this forecloses nothing.

linkml_model/model/schema/meta.yaml
```diff
@@ -650,6 +650,29 @@ slots:
...
+    domain: enum_expression
+    range: boolean
+    description: >-
+      Whether the enumeration is open, meaning that values outside of the set of
+      permissible values are allowed. If true, the enumeration is open and data values
+      that are not in the set of permissible values are permitted (though they SHOULD be
+      drawn from the permissible values if available). If false (runtime default), the
+      enumeration is closed and data values MUST be drawn from the set of permissible values.
+    ifabsent: false
```
amc-corey-cox: Drop `ifabsent: false`. It materializes `False` on every enum and the dumper emits it, so all schemas grow an `is_open: false` line. No other boolean slot in the metamodel uses `ifabsent`.

linkml_model/model/schema/meta.yaml
```diff
@@ -650,6 +650,29 @@ slots:
...
+    comments:
+      - An open enumeration is analogous to the `allow-other="yes"` construct in the OSCAL Metamodel.
+      - When an enum is open, the permissible values are advisory - they document the recommended
```
amc-corey-cox: This contradicts the description above. Open means the value set isn't exhaustive, not that other constraints vanish — a slot `pattern` should still hold. The "maps to a plain string" bit is a JSON Schema lowering detail; leave that to linkml#3712 and remove the reference to JSON Schema.

linkml_model/model/schema/meta.yaml
```diff
@@ -650,6 +650,29 @@ slots:
...
+        or expected values but do not constrain the range. In JSON Schema this maps to a plain string.
+    see_also:
+      - https://pages.nist.gov/metaschema/specification/syntax/constraints/#allow-other
+      - https://github.com/linkml/linkml/issues/127
+      - https://github.com/linkml/linkml/pull/3712
+    in_subset:
+      - SpecificationSubset
+      # BasicSubset
```
amc-corey-cox: Drop the commented-out `# BasicSubset`.

linkml_model/model/schema/meta.yaml
```diff
@@ -650,6 +650,29 @@ slots:
...
+    status: testing
```
amc-corey-cox, line 672: Worth a `tests/input/examples/schema_definition-open-enums.yaml` so `make test-examples` covers it.

amc-corey-cox (changes requested):
The shape looks right to me. is_open as a boolean fits the metamodel: it mirrors id_prefixes_are_closed — one slot builds the value set, another says whether it's exhaustive. is_open: true just demotes the permissible values from MUST to SHOULD; it doesn't collapse the range.

@sneakers-the-rat on binary-vs-trinary: I don't think enums need the all/none/some shape that extra_slots has. extra_slots had to be trinary because it has no set-construction of its own. Enums already do that via permissible_values/reachable_from/matches/include/minus. And per-value constraints already work: a slot pattern holds independently of enum membership, so "extras must be well-formed" needs nothing new.

If a real case for open_to: <expr> shows up, that will need more definition than we're ready to do here and we should design around the live case.

noelmcloughlin:
Changes done. Raised https://github.com/linkml/linkml-model/pull/265 and https://github.com/linkml/linkml-model/pull/266 due to test failures.

linkml_model/model/schema/meta.yaml
amc-corey-cox: Still looks right. Nothing to change here - leaving this as the record of why a boolean fits.

linkml_model/model/schema/meta.yaml
amc-corey-cox: Fixed. `ifabsent` is gone and generated `meta.py` now has `is_open: Optional[Union[bool, Bool]] = None`, so absent stays absent.

linkml_model/model/schema/meta.yaml
amc-corey-cox: Fixed. The comment now frames open as "the permissible value set is not exhaustive" and the JSON Schema lowering detail is gone; linkml#3712 stays as a `see_also`, which is the right place for it.

linkml_model/model/schema/meta.yaml
amc-corey-cox: Fixed. `in_subset` is just `SpecificationSubset` now, with `status: testing`.

linkml_model/model/schema/meta.yaml
amc-corey-cox: Done, and then some. `schema_definition-open-enums.yaml` covers named enums, anonymous `enum_range`, `slot_usage` narrowing, dynamic `reachable_from`, and composed `inherits`/`include`/`minus` - the Makefile glob picks it up. `tests/test_open_enums.py` round-trips it through the generated metamodel on top of that.

amc-corey-cox (dismissed):
This shouldn't merge ahead of #265 — the pre-commit config currently rides in on a merge commit and deserves its own review. Otherwise all five of my comments are addressed: `ifabsent` is gone, the description no longer contradicts itself, and the example plus round-trip tests go further than I asked for.

sneakers-the-rat:
> on binary-vs-trinary

OK cool, makes sense to me, then in general we are not going for naming consistency for openness across the different places that we want to declare it? We should probably make some kind of grouping/subset for openness terms then, or else make sure they are documented together. E.g. we might sprout an `extra_classes` at the schema level, and then we have 

- `is_open` - on enums
- `extra_slots` - on classes
- `extra_classes` - maybe, on schema (name undecided)
- `open_world` - I'm actually not quite sure how this works, but its confusable by name and should be differentiated

I think just having a listing like that on each that refers to the others as part of a set in the docs would be enough. Main concern I have is being confusing/hard to discover if similar concepts are named differently.

noelmcloughlin:
Fixed (many) conflicts, squashed my commits.

amc-corey-cox (approved):
This looks good to me.

amc-corey-cox:
@noelmcloughlin - I'll get this merged right after our next release. I'll keep it de-conflicted and get it in as soon as I can.

PR #262, opened 2026-07-07
Add "type_designator_mapping" slot.

gouttegd:
Add a new `type_designator_mapping` slot.

That slot is intended to be used in a slot_definition for a slot that is set as the type designator slot for the class it belongs to.

It allows to define custom mappings to match a type designator value with the type to be designated. Such mappings can be used to provide more ways to reference a given class in a type designator value (beyond the default way of using the class name, URI, or CURIE).

This implements the idea evoked in https://github.com/linkml/linkml/issues/3701.

### DIFF
--- a/linkml_model/model/schema/meta.yaml
+++ b/linkml_model/model/schema/meta.yaml
@@ -1934,6 +1934,39 @@ slots:
       - SpecificationSubset
     see_also:
       - https://linkml.io/linkml/schemas/type-designators.html
+      - linkml:type_designator_mapping
+
+  type_designator_mapping:
+    domain: slot_definition
+    range: designated_type_mapping
+    multivalued: true
+    inlined: true
+    inlined_as_list: false
+    inherited: true
+    description: >-
+      Additional mappings between type designator values and instantiation types.
+
+      Mappings defined here are taken into account after a type designator value failed to match the
+      name, URI, or CURIE of any known subclass of the class carrying the type designator slot.
+    comments:
+      - This slot only makes sense for a slot that is a type designator slot.
+    examples:
+      - object:
+          foo: FooClass
+          bar: BarClass
+          "*": FallbackClass
+        description: >-
+          Assuming FooClass, BarClass, and FallbackClass all refers to classes that are compatible
+          with the base class carrying the type designator slot, this example would allow to use
+          'foo' to designate an instance of FooClass and 'bar' to designate an instance of BarClass;
+          in addition, any unrecognised type name is to be interpreted as designating an instance of
+          FallbackClass.
+    in_subset:
+      - SpecificationSubset
+    see_also:
+      - https://linkml.io/linkml/schemas/type-designators.html
+      - linkml:designates_type
+      - linkml:designated_type_mapping
 
   alias:
     rank: 6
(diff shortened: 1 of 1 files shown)


gouttegd:
@cmungall This proposal has the following differences compared to what we discussed in https://github.com/linkml/linkml/issues/3701:

(A) The keys in the `type_designator_mapping` map are expected to be simple, “shell-like” patterns (where the only special characters are `*`, standing for anything, and `?`, standing for any single character), rather than full-fledged regular expressions.

So for example, the map entry that would act as the ultimate fallback (recognising anything) would be just `*`, instead of `.*`.

This is because I am not convinced we need the expressivity of full regular expressions here – this seems “overkill” to me.

However I don’t have too much of a strong opinion here. If you (or anyone else for that matter) believe regular expressions would really be beneficial here, that’s fine with me, happy to change.

(B) I introduced the notion of “longest match”, in that the designated type to be effectively used has to be found by searching for the _longest_ matching entry in the `type_designator_mapping` map.

This is because the `type_designator_map` is serialised as a _dictionary_. Because of that, we cannot rely on the order of keys to avoid a premature match.

Consider the example coming from the https://github.com/linkml/linkml/issues/3701 ticket:

```yaml
type_designator_mapping:
  TC: TargetingComponent
  LC: LoadingComponent
  "*": FrobnicatorComponent
```

There is no intrinsic guarantee that the `*` key will be tried last! If the value of the type designator is, say, `LC` and the `*` key is tried _before_ the `LC` key, then `LC` will be found to match the `*` entry (because it does, since `*` matches anything), and the object will be assumed to be a `FrobnicatorComponent` instead of the intended `LoadingComponent`.

The “longest match” rule avoids this situation. `LC` would match the key `LC`, and because it’s a longer (more precise) pattern, it would take precedence over the shorter `*` pattern.

linkml_model/model/schema/meta.yaml
```diff
@@ -1934,6 +1934,39 @@ slots:
...
+
+      Mappings defined here are taken into account after a type designator value failed to match the
+      name, URI, or CURIE of any known subclass of the class carrying the type designator slot.
+    comments:
+      - This slot only makes sense for a slot that is a type designator slot.
+    examples:
+      - object:
+          foo: FooClass
+          bar: BarClass
+          "*": FallbackClass
```
Ostrzyciel, line 1957: If we want to guarantee a specific apply order, this should be a list of mappings, or something similar, e.g.:

```yaml
- object:
    - foo: FooClass
    - bar: BarClass
    - "*": FallbackClass
```

The longest-match precedence rule would require you to evaluate all possibilities, I think, which would not be very efficient. There also cases where you have two matchers with equal lengths, then it would be matched randomly.

In languages with pattern matching, matchers have a specific evaluation order so that they are more predictable.

linkml_model/model/schema/meta.yaml
```diff
@@ -3332,6 +3365,33 @@ classes:
...
+  designated_type_mapping:
+    description: >-
+      A custom mapping between a pattern and a class definition.
+    attributes:
+      pattern:
+        description: >-
+          The pattern to map to a class definition.
+
+          This is a shell-like pattern, where '*' represents any substring (including an
+          empty string) and '?' represents any character.
```
Ostrzyciel, line 3377: Could we define a grammar for these patterns? A little bit of BNF helps tremendously with implementations. This, plus an explanation what each production expands to.

Ostrzyciel, line 3377: Are there are any other pieces of the spec / metamodel that would benefit from such patterns? E.g., in linkml-map? I'd very much support unifying the different expression languages to make it easier to maintain implementations. LinkML already has at least 2 languages that I'm aware of, this would be a 3rd :)

This is not an objection (I like the feature!), but rather an ask to look at this a bit more holistically.

gouttegd:
Following [the discussion](https://github.com/linkml/linkml/issues/3701#issuecomment-5428760611) in today’s dev meeting, closing. The feature requested in linkml/linkml#3701 will be implemented using another approach.

PR #263, opened 2026-07-14
fix: `status`'s range is `string` instead of `uriorcurie`

### LINKED ISSUE #3760: metamodel status slot has wrong `slot_uri` (`bibo:status`) and mixed semantics — term maturity (`vs:term_status` strings) vs. document lifecycle (`bibo:DocumentStatus` URIs) need separate treatment
**Describe the bug**

I'm working on adding to `expand_curie()` strict input validation: either a CURIE (that gets expanded) or a URI (which is left untouched). And such validation has raised this issue.

The metamodel's `status` slot has conflicting semantics, an incorrect `slot_uri`, and an incorrect `range` for its actual usage.

What the metamodel declares:
```yaml
status:
  domain: element
  range: uriorcurie
  slot_uri: bibo:status
  examples:
    - value: "bibo:draft"
```

**Problem 1** - wrong `slot_uri`

`bibo:status` (<http://purl.org/ontology/bibo/status>) is an OWL object property
whose `range` is `bibo:DocumentStatus` - a class of named individuals representing
document publication lifecycle states (`accepted`, `draft`, `published`, `unpublished`,
…).  Its individuals live at http://purl.org/ontology/bibo/status/<name>, so even
the example value `bibo:draft` is wrong: the correct URI would require a separate
prefix such as `bibo_status: http://purl.org/ontology/bibo/status/`.

**Problem 2** - wrong `range` for actual usage

The values actually used throughout `meta.yaml` and `array.yaml` are the bare
strings `testing` (30+ occurrences, like in [meta.yaml](https://github.com/linkml/linkml-model/blob/986c7881bae48055d3dbd9106205bc86c5215df4/linkml_model/model/schema/meta.yaml#L862) and [array.yaml](https://github.com/linkml/linkml-model/blob/986c7881bae48055d3dbd9106205bc86c5215df4/linkml_model/model/schema/array.yaml#L13)) and `unstable` (2 occurrences, like in [meta.yaml](https://github.com/linkml/linkml-model/blob/986c7881bae48055d3dbd9106205bc86c5215df4/tests/input/meta.yaml#L1373)).  These are
not URIs or CURIEs - they are plain-string values probably from the W3C SemWeb Vocab
Status vocabulary (`vs:term_status`,
<http://www.w3.org/2003/06/sw-vocab-status/ns#term_status>), which explicitly
defines its values as literals: `"stable"`, `"testing"`, `"unstable"`, `"archaic"`.

**Problem 3** - mixed semantics: terms vs. documents

The `status` slot applies to element, the root class for both schema terms
(classes, slots, types, enums) and schema documents.  The two use cases have
incompatible semantics:
| Subject | Vocabulary | Property | Range |
|---|---|---|---|
| Vocabulary terms | W3C SemWeb Vocab Status | `vs:term_status` | string literal |
| Documents / schemas | bibo | `bibo:status` | `bibo:DocumentStatus` URI |

A single `slot` with a single `slot_uri` cannot correctly represent both.

*Practical consequence*

Any tool that calls `expand_curie("testing")` on a `status` slot value would raise
`ValueError: 'testing' is neither a CURIE nor a full URI` on a strict URI/CURIE validation, because `"testing"` is
a bare word with no registered prefix.

*Proposed fix*
Declare a controlled vocabulary of status values in the LinkML ontology
(https://w3id.org/linkml/), covering both term-maturity and document-lifecycle
concepts.  Then either:
- **Option A** - single slot, linkml-native URI:
   drop `slot_uri: bibo:status`, use `linkml:status` as the native property with `range: uriorcurie`, and express
all values as linkml: CURIEs (e.g. `linkml:testing`, `linkml:draft`).
- **Option B** - two slots:
   split into `term_status` (`slot_uri: vs:term_status`, `range: string`, `domain: term elements`) and `status` (`slot_uri: bibo:status`, `range: uriorcurie`, `domain: schema/document elements`), with `bibo:DocumentStatus` individuals or linkml-owned equivalents as values for the latter.
- **Option C** - change `range` from `uriorcurie` to `string`:
   The `status` slot applies to both vocabulary terms and schema documents, and in practice its values are always plain strings (e.g. `testing`, `unstable`, `draft`). Rather than attempting to enforce URI semantics — which the current usage already violates — the `range` is corrected to `string` to accurately reflect how the slot is used. `bibo:status` and `vs:term_status` can become `linkml:close_mappings` of the slot; consumers that need stricter document-lifecycle semantics should use a dedicated slot. This change is non-breaking: all existing `status:` values in schemas are already valid strings.

Options A and B are breaking changes for existing schemas that use `status:` on terms with bare-word values. Only Option C is not a breaking change, with the above mentioned drawbacks.

Silvanoc:
The slot `status` declares it's range to be `uriorcurie`, but in fact bare strings like `testing` or `unstable` are being used in the metamodel itself (more details in the below linked issue). This patch aligns the specification with its current use.

Fixing it to remain `uriorcurie` would require multiple changes with bigger impact than this change. They might get implemented in the future.

Fixes: https://github.com/linkml/linkml/issues/3760

### DIFF
--- a/linkml_model/model/schema/meta.yaml
+++ b/linkml_model/model/schema/meta.yaml
@@ -434,11 +434,13 @@ slots:
     aliases:
       - workflow status
     domain: element
-    range: uriorcurie
+    range: string
     description: status of the element
-    slot_uri: bibo:status
+    close_mappings:
+      - bibo:status
     examples:
-      - value: "bibo:draft"
+      - value: "testing"
+      - value: "unstable"
     see_also:
       - https://www.hl7.org/fhir/valueset-publication-status.html  ## Draft, Active, Retired, Unknown
       - https://www.hl7.org/fhir/versions.html#std-process  ## Draft, Trial Use, Normative, Informative, Deprecated


matentzn (commented):
@Silvanoc can you redo the PR with just the model changes? I will do the generation of files on a slightly different cadance (it also makes it clearer which files have actially changed other than linkml_model/model/schema/meta.yaml)

Silvanoc:
> @Silvanoc can you redo the PR with just the model changes? I will do the generation of files on a slightly different cadance (it also makes it clearer which files have actially changed other than linkml_model/model/schema/meta.yaml)

@matentzn sure! Guess why I have two commits with exactly that split 😉 I'll drop the 2nd one and done.

linkml_model/model/schema/meta.yaml
```diff
@@ -434,11 +434,13 @@ slots:
     aliases:
       - workflow status
     domain: element
-    range: uriorcurie
+    range: string
```
noelmcloughlin, line 437: Could we test this change against OWL generator?  

While investigating https://github.com/linkml/linkml/issues/3803 I tried this patch but found `OwlSchemaGenerator.add_metadata()` is silently dropping status annotation from OWL output due to this `owlgen.py` conditional

 ```
 if metaslot_curie.startswith("linkml:"):
     # only mapped properties
     continue
```

This skips metaslots whose URI falls back to the native linkml: namespace - and after deleting `slot_uri: bibo:status`, the slot's URI becomes `linkml:status`, and every element's status value vanishes from the generated OWL.

https://github.com/linkml/linkml/blob/8e443f15faa4673acc3345923ea84199f73e7f50/packages/linkml/src/linkml/generators/owlgen.py#L379

Comment (Silvanoc) [reply linkml_model/model/schema/meta.yaml:437]: @noelmcloughlin it might be related with issue https://github.com/linkml/linkml/issues/3752. Please give the attached PR a try.

Comment (noelmcloughlin) [reply linkml_model/model/schema/meta.yaml:437]: Seems different bug... let me clarify

this PR alone is problematic.  After removing `slot_uri: bibo:status`, slot's URI becomes `linkml:status`, and every element (ArrayExpression, apply_to, range_expression, etc.) status value vanishes from generated OWL (on my branch, I confirmed zero "testing" occurrences in my output, versus ~32  bibo:testing triples using alternative `uriocurie` approach).  So a sibling PR for`owlgen.py` is needed to address fallout

Comment (Silvanoc) [reply linkml_model/model/schema/meta.yaml:437]: Hmm, this PR is not changing `status` itself, only its `range`. Since you have tried it out, I assume that the explanation is not precise, but the problematic effect exists.

I'll switch this PR to `draft` until I've clarified the situation and, it needed, provided the required sibling PR for `owlgen.py`.

Comment (Silvanoc) [reply linkml_model/model/schema/meta.yaml:437]: @noelmcloughlin I've realized this issue trying to make `SchemaView.expand_curie` strict (trying to expand something that is neither a URI -no expansion needed- nor a CURIE -can be expanded) on PR https://github.com/linkml/linkml/pull/3805 (specifically commit https://github.com/linkml/linkml/pull/3805/changes/85fb0c9534f62d56c173cf3c59fb820a2f446fa2). Apparently this and that PRs are somehow coupled together.

Comment (noelmcloughlin) [reply linkml_model/model/schema/meta.yaml:437]: yes, three lines in `owlgen.py` are coupled to this PR 
```
if metaslot_curie.startswith("linkml:"):
    # only mapped properties
    continue
```

Comment (Silvanoc) [reply linkml_model/model/schema/meta.yaml:437]: @noelmcloughlin Can you give me the link to a schema to reproduce the issue?

With the `kitchen_sink.yaml` schema I cannot see any difference on how [this spot](https://github.com/linkml/linkml/blob/8e443f15faa4673acc3345923ea84199f73e7f50/packages/linkml/src/linkml/generators/owlgen.py#L379) behaves after the change in the model. I'm printing out `metaslot_curie` and saving the output with and without the model change and I get exactly the same values.

Comment (Silvanoc) [reply linkml_model/model/schema/meta.yaml:437]: The owlgen spot is about the slot URI, not about the value (what this patch changes), therefore I'm unsure if the diagnose is correct.

File: linkml_model/model/schema/meta.yaml [hunk 2]
```diff
@@ -434,11 +434,13 @@ slots:
...
     description: status of the element
-    slot_uri: bibo:status
+    close_mappings:
+      - bibo:status
```
kevinschaper: `range: uriorcurie` → `string` is the real fix here and it works. But replacing `slot_uri: bibo:status` with `close_mappings` in the same patch removes every `status` assertion from the generated artifacts.

`slot_uri` is the predicate the generators actually write with. `close_mappings` is documentation — nothing emits it. So without a `slot_uri` the generator still knows each element's status but has no predicate to write it under, and drops the triples.

Measured against linkml `main` (`1e9c7ba6a`), `gen-owl` over `meta.yaml`, fresh process per run, `PYTHONHASHSEED ∈ {0,1,42}`:

| variant | canonicalization warning | status assertions | bytes |
|---|---|---|---|
| current `main` | 1 | 32, as `<testing>` / `<unstable>` | 205,615 |
| **this PR as written** | 0 | **0** | 191,076 |
| `range: string`, keep `slot_uri` | 0 | **32, as `"testing"` / `"unstable"`** | 191,847 |
| keep `uriorcurie`, use `close_mappings` | 0 | 0 | 191,076 |

Rows 2 and 4 come out byte-identical, which pins the loss on the `slot_uri` removal rather than on the range change.

Keeping `slot_uri` produces:

```turtle
bibo:status "testing"     # ×30
bibo:status "unstable"    # ×2
```

Valid RDF, all 32 assertions preserved, and pyoxigraph can now parse the graph — so `canonicalize_rdf_graph` performs real RDFC-1.0 instead of falling back to rdflib. That resolves linkml/linkml#3516 bullet 3 completely.

```suggestion
    range: string
    description: status of the element
    slot_uri: bibo:status
```

On whether `bibo:status` is the semantically right predicate at all — that's Problems 1 and 3 in #3760, and Options A/B are parked for it. It's a genuine question, but `bibo:status "testing"` is valid RDF and strictly better than the relative-IRI `<testing>` shipping today, so it doesn't need to block this fix.

linkml_model/model/schema/meta.yaml
Silvanoc: Reminder @noelmcloughlin: I'm missing a schema to reproduce the issue.

noelmcloughlin: Hi,
I will resolve this comment as I searched my local notes for this issue and coul not find it, so there is no evidence i can reach for, apologies

Silvanoc: I would then say that there is no blocker anymore... still careful reviewing and testing will be needed to avoid breaking downstream.

linkml_model/model/schema/meta.yaml
kevinschaper: `range: uriorcurie` → `string` is the fix and it works. But replacing `slot_uri: bibo:status` with `close_mappings` in the same patch removes every `status` assertion from the generated OWL.

`slot_uri` is the predicate the generators write with. `close_mappings` is documentation — nothing emits it. Without a `slot_uri` the generator still knows each element's status but has no predicate to write it under, so it drops the triples.

Measured by generating this repo's `meta.yaml` (`main` @ `35c91fb0`) and varying only the `status` slot:

| variant | canonicalization warning | relative IRIs | status assertions | bytes |
|---|---|---|---|---|
| current `main` | 1 | 32 | 0 | 205,620 |
| **this PR as written** | 0 | 0 | **0** | 191,001 |
| `range: string`, keep `slot_uri` | 0 | 0 | **32** | 191,852 |

Both fixed variants clear the relative IRIs and let pyoxigraph parse the graph, so `canonicalize_rdf_graph` performs real RDFC-1.0 instead of falling back to rdflib. The difference is whether the statements survive: keeping `slot_uri` produces

```turtle
bibo:status "testing"     # ×30
bibo:status "unstable"    # ×2
```

while the PR as written produces nothing at all for `status`, losing ~14.6 KB of assertions.

```suggestion
    range: string
    description: status of the element
    slot_uri: bibo:status
```

On whether `bibo:status` is the semantically right predicate in the first place — that's Problems 1 and 3 in linkml/linkml#3760, and Options A and B are parked there for it. A fair question, but `bibo:status "testing"` is valid RDF and strictly better than the relative-IRI form shipping today, so it needn't block this fix.

The updated `examples:` (`testing` / `unstable`) are right either way — the old `"bibo:draft"` example never matched actual usage.

kevinschaper: `range: string` is correct. Demoting `slot_uri` to `close_mappings` is the problem: `slot_uri` is the predicate generators write with, `close_mappings` is documentation nothing emits. Without it the generator knows each status but has no predicate for it, so it emits nothing.

Generating this repo's `meta.yaml` on `main` (`35c91fb0`), varying only this slot:

| variant | relative IRIs | status assertions | bytes |
|---|---|---|---|
| `main` | 32 | 0 | 205,620 |
| **this PR** | 0 | **0** | 191,001 |
| range only | 0 | **32** | 191,852 |

Keeping `slot_uri` gives `bibo:status "testing"` ×30 / `"unstable"` ×2 — valid RDF, pyoxigraph parses it, RDFC-1.0 runs.

```suggestion
    range: string
    description: status of the element
    slot_uri: bibo:status
```

Whether `bibo:status` is the right predicate at all is linkml/linkml#3760 Problems 1 and 3 — Options A/B are parked there and needn't block this. The new `examples:` are right either way; `"bibo:draft"` never matched real usage.

kevinschaper: This drops all 32 status assertions from the generated OWL.

Generating this repo's `meta.yaml` with the change applied (linkml `origin/main`):

```
bibo:status "testing" / "unstable"    32 → 0
linkml:status assertions              0 → 0
linkml:status property declaration    still emitted
meta.owl.ttl                     205,620 → 191,001 bytes
```

`slot_uri` is the predicate assertions are written under; `close_mappings` is documentation and nothing emits it. The property stays declared, but no element carries a status under any predicate.

#3760's Option C prescribes exactly this step and calls it non-breaking — which holds for schema authors, since existing `status:` values remain valid strings, but not for the published artifacts.

If `bibo:status` is the wrong predicate (Problem 1 in #3760), Option C needs to name the replacement — Option A's `linkml:status` — because removing it leaves nothing to assert under.

kevinschaper: This drops all 32 status assertions from the generated OWL.

Generating this repo's `meta.yaml` with the change applied (linkml `origin/main`):

```
bibo:status "testing" / "unstable"    32 → 0
linkml:status assertions              0 → 0
linkml:status property declaration    still emitted
meta.owl.ttl                     205,620 → 191,001 bytes
```

`slot_uri` is the predicate assertions are written under; `close_mappings` functions as documentation only. The property stays declared, but no element carries a status under any predicate.

#3760's Option C prescribes exactly this step and calls it non-breaking — which holds for schema authors, since existing `status:` values remain valid strings, but not for the published artifacts.

If `bibo:status` is the wrong predicate (Problem 1 in #3760), Option C needs to name the replacement — Option A's `linkml:status` — because removing it leaves nothing to assert under.

kevinschaper: This drops all 32 per-element status assertions from the generated OWL.

Generating this repo's `meta.yaml` with the change applied (linkml `origin/main`), varying only this slot:

```
                status assertions   relative IRIs   skos:closeMatch   bytes
main                    0                32               0          205,615
this PR                 0                 0               1          191,076
range only             32                 0               0          191,847
```

`close_mappings` is still exported — it becomes `skos:closeMatch bibo:status` on the `linkml:status` declaration. What it can't do is carry the values: `slot_uri` is the predicate per-element assertions are written under, so removing it leaves the property declared with a mapping annotation but no element actually carrying a status.

#3760's Option C prescribes exactly this step and calls it non-breaking — true for schema authors, since existing `status:` values remain valid strings, but not for the published artifacts.

If `bibo:status` is the wrong predicate (Problem 1 in #3760), Option C needs to name the replacement — Option A's `linkml:status` — because removing it leaves nothing to assert under.

kevinschaper (approved):
This looks good to me. I went down a bit of a rabbit hole of my agent believing that downstream, artifacts would leave status out because of the change from slot_uri to close_mappings, and after testing it, the only thing I could confirm was that the owl representation of the schema will change, and I think that's what is expected.

noelmcloughlin (approved):
Lgtm, thanks

PR #265, opened 2026-07-27
test(pre-commit): exclude gen-project artefact failures we cannot control

noelmcloughlin:
This PR is to stop CI failures for `gen-project` generated output (which `linkml-model` does not control):

Excluded quality hooks:

- `end-of-file-fixer`
- `trailing-whitespaces`

Excluded generated content:

- excel/ graphql/ jsonld/ jsonschema/ owl/ prefixmap/ protobuf/ rdf/ shacl/ shex/ sqlddl/ sqlschema/
- annotations array datasets extensions mappings meta types units validation .py

See: https://github.com/linkml/linkml-model/actions/runs/30282095966/job/90030765962

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,14 +1,26 @@
 default_language_version:
   python: python3
 
+# Note: the `exclude` patterns below cover `make gen-project` outputs written
+# into linkml_model/ which we do not control. Artefacts copied verbatim from the
+# LinkML generators, which emit trailing whitespace and inconsistent final
+# newlines (e.g. the bare "// " comment line in protobuf/meta.proto). Native
+# sources stay covered: model/schema/, model/docs/, linkml_files.py, etc.
+
 repos:
 - repo: https://github.com/pre-commit/pre-commit-hooks
   rev: v6.0.0
   hooks:
   - id: end-of-file-fixer
+    exclude: &generated-artefacts |
+      (?x)^linkml_model/(
+        (excel|graphql|jsonld|jsonschema|owl|prefixmap|protobuf|rdf|shacl|shex|sqlddl|sqlschema)/
+        |(annotations|array|datasets|extensions|mappings|meta|types|units|validation)\.py$
+      )
   - id: check-yaml
   - id: check-toml
   - id: trailing-whitespace
+    exclude: *generated-artefacts
 
 - repo: https://github.com/codespell-project/codespell
   rev: v2.4.1


.pre-commit-config.yaml
```diff
@@ -1,14 +1,31 @@
 default_language_version:
   python: python3
 
+# Note: the `exclude` patterns below cover `make gen-project` outputs written
+# into linkml_model/ which we do not control. Artefacts copied verbatim from the
+# LinkML generators, which emit trailing whitespace and inconsistent final
+# newlines (e.g. the bare "// " comment line in protobuf/meta.proto). Native
+# sources stay covered: model/schema/, model/docs/, linkml_files.py, etc.
+# Note: Keep the two exclude patterns in sync.
```
matentzn: Cant we us a YAML merge anchor (&gen_artefacts / *gen_artefacts) to remove the duplication and the sync fragility entirely?

.pre-commit-config.yaml
```diff
@@ -1,14 +1,31 @@
...
+
 repos:
 - repo: https://github.com/pre-commit/pre-commit-hooks
   rev: v6.0.0
   hooks:
   - id: end-of-file-fixer
+    exclude: |
+      (?x)^linkml_model/(
+        (excel|graphql|jsonld|jsonschema|owl|prefixmap|protobuf|rdf|shacl|shex|sqlddl|sqlschema)/
+        |(annotations|array|datasets|extensions|mappings|meta|types|units|validation)\.py$
```
matentzn, line 18: Hand edited list will silently drift as new generators are added. Not a blocker for me, just noting this in case you can think of a guard.

.pre-commit-config.yaml
noelmcloughlin: Done. I was testing this previously and had issue - but fresh attempt works and `pre-commit` runs okay.

.pre-commit-config.yaml
noelmcloughlin: Nothing comes to mind with current layout - **include vs exclude** list is needed

matentzn: np. I was thinking maybe its possible to pull a generator list from somewhere and expand it into a string somehow, but thats overkill. I will approve.

amc-corey-cox: @matentzn - I played around with options for this but none were any more clear or less fragile than just keeping the hand edited list. The better long-term solution is to move this down in to the linkml vendoring and then remove it altogether. But we need to slow walk that because of the model-linkml coupling.

amc-corey-cox (commented):
This looks good to me.

amc-corey-cox (approved):

kevinschaper (approved):

PR #266, opened 2026-07-27
fix(tests): nest structured annotation payload under `value`

noelmcloughlin:
Fixes #https://github.com/linkml/linkml/issues/3809

- Nesting the payload under `value` (range AnyValue) preserves the structure and lets the file load
- Hard fail `tests/test_input_against_model.py` when counted failures

### DIFF
--- a/tests/input/examples/schema_definition-native-array-1.yaml
+++ b/tests/input/examples/schema_definition-native-array-1.yaml
@@ -20,13 +20,17 @@ classes:
   TemperatureDataset:
     tree_root: true
     annotations:
+      # An annotation is a tag/value pair. A structured payload must be nested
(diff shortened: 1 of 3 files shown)


matentzn (commented):
The change is good, but why dont we take this opportunity to hard fail in `tests/test_input_against_model.py`? Like this it also wont happen again... If I know go and add the same wrong modelling to a new test it will just be ignored again..

```
self.assertEqual(nfailures, 0, f"{nfailures} input file(s) failed to load")
```

noelmcloughlin:
Added

amc-corey-cox:
@noelmcloughlin - I'll get this merged right after our next release. I'll keep it de-conflicted and get it in as soon as I can.

amc-corey-cox (approved):
This looks good to me.

PR #268, opened 2026-08-04
Fix spec links and duplicate generated docs

Ostrzyciel:
- Fix broken link in specification index.md
- Write the links with the .md extension so that mkdocs can catch broken links
- Fix `make gen-doc` generating increasingly nested `linkml_model` directories in `docs/`, which causes a mess in local development
- Add an mkdocs build to PR CI so that broken internal links are caught in CI

### DIFF
--- a/.github/workflows/build-publish-documentation.yaml
+++ b/.github/workflows/build-publish-documentation.yaml
@@ -44,6 +44,9 @@ jobs:
       - name: Generate markdown docs from schema
         run: uv run make gen-doc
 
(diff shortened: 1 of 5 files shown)


matentzn (approved):
Amazing thanks @Ostrzyciel! 

One small FYI; when reviewing your PR I noticed (unrelated to your PR) that `07codegen.md` exists in `specification/`, but is not mentioned in mkdocs or index.md. Since you are already in cleaning mode..

PR #270, opened 2026-08-06
Mixin inheritance clarification

### LINKED ISSUE #3849: Specification: Clarify mixin inheritance structural conformance rules
**Describe the bug**
Currently, the specification states that “For any parent p of c, if p.mixin is True, then c.mixin SHOULD be True” [Derived Schemas - LinkML Model](https://linkml.io/linkml-model/latest/docs/specification/04derived-schemas/#classdefinition-structural-conformance-rules). I believe “parent” in this context refers to a direct `is_a` parent OR `mixins` parents. This would imply that any class that has any mixin would have to be a mixin as well. This seems unintended, and the rule should specify that the parent should be a direct `is_a` parent instead.

**Please provide a schema (and if applicable, a data file) that replicates the issue**
This seems to be used in the metamodel: `EnumExpression` is not declared as a mixin, but is used as one in `EnumDefinition`, because `EnumExpression` has an `Expression` direct parent that is declared as a mixin.

niegrzybkowski:
Closes https://github.com/linkml/linkml/issues/3849

### DIFF
--- a/linkml_model/model/docs/specification/04derived-schemas.md
+++ b/linkml_model/model/docs/specification/04derived-schemas.md
@@ -488,7 +488,7 @@ Each `c` in *m<sup>D</sup>*`.classes` must conform to the rules below:
 - `c` may have parent ClassDefinitions defined via `c.is_a` and `c.mixins`
     - the value of `c.is_a` must be a ClassDefinitionReference
     - the value of `c.mixins` must be a collection of ClassDefinitonReferences
-   - For any parent `p` of `c`, if `p.mixin` is True, then `c.mixin` SHOULD be True
+    - for any direct (`is_a`) ancestor `a` of `c`, if `a.mixin` is True, then `c.mixin` SHOULD be True
 - `c` includes additional rules in `c.rules` and `c.classificiation_rules`
 - `c` may have any number of additional slot-value assignments consistent with the validation rules provided here with the metamodel `MM`
 
@@ -505,7 +505,7 @@ Each `s` in *m<sup>D</sup>*`.slots` must conform to the rules below:
 - `s` may have parent SlotDefinitions defined via `s.is_a` and `s.mixins`
     - the value of `s.is_a` must be a **SlotDefinitionReference**
     - the value of `s.mixins` must be a collection of **SlotDefinitionReference**s
-    - For any parent `p` of `s`, if `p.mixin` is True, then `s.mixin` SHOULD be True
+    - for any direct (`is_a`) ancestor `a` of `s`, if `a.mixin` is True, then `s.mixin` SHOULD be True
 - `s` may have any number of additional slot-value assignments consistent with the validation rules provided here with the metamodel `MM`
 
 ### TypeDefinition Structural Conformance Rules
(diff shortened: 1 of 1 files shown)


linkml_model/model/docs/specification/04derived-schemas.md
```diff
@@ -488,7 +488,7 @@ Each `c` in *m<sup>D</sup>*`.classes` must conform to the rules below:
 - `c` may have parent ClassDefinitions defined via `c.is_a` and `c.mixins`
     - the value of `c.is_a` must be a ClassDefinitionReference
     - the value of `c.mixins` must be a collection of ClassDefinitonReferences
-   - For any parent `p` of `c`, if `p.mixin` is True, then `c.mixin` SHOULD be True
+    - for any direct (`is_a`) ancestor `a` of `c`, if `a.mixin` is True, then `c.mixin` SHOULD be True
```
matentzn, line 491: "direct ancestor" is a bit of a contradiction, so better use the word "parent" (I dont mind the redundant "direct"):

```suggestion
    - for any direct `is_a` parent `a` of `c`, if `a.mixin` is True, then `c.mixin` SHOULD be True
```

linkml_model/model/docs/specification/04derived-schemas.md
```diff
@@ -505,7 +505,7 @@ Each `s` in *m<sup>D</sup>*`.slots` must conform to the rules below:
 - `s` may have parent SlotDefinitions defined via `s.is_a` and `s.mixins`
     - the value of `s.is_a` must be a **SlotDefinitionReference**
     - the value of `s.mixins` must be a collection of **SlotDefinitionReference**s
-    - For any parent `p` of `s`, if `p.mixin` is True, then `s.mixin` SHOULD be True
+    - for any direct (`is_a`) ancestor `a` of `s`, if `a.mixin` is True, then `s.mixin` SHOULD be True
```
matentzn, line 508:
```suggestion
    - for any direct `is_a` parent `a` of `s`, if `a.mixin` is True, then `s.mixin` SHOULD be True
```

matentzn (dismissed):
I think this is correct.

EDIT: withdrew some minor concerns, misread the issue.

gouttegd (changes requested):
(Asked to review by @matentzn )

(i) I agree with the assessment in the original ticket that the current phrasing in the spec is wrong, as it would indeed imply that any class that uses a mixin (has a `mixins` parent) should itself be a mixin, which wouldn’t make any sense.

(ii) I agree with @matentzn ’s objection regarding the use of “direct ancestor” as a way to refer to a “`is_a` parent”. This is not how most people would understand such a term. In the absence of any established terminology to distinguish between “parent through a `is_a` relationship” and “parent through a `mixins` relationship”, better to always be explicit and to always talk about “`is_a` parent” and “`mixins` parent”.

So overall, agreeing with the general idea, but requesting the changes suggested by @matentzn.

niegrzybkowski:
Applied the suggestions, also renamed the parent variable back to `p`. 

@matentzn  @gouttegd Good point, requesting re-review

matentzn (approved):

gouttegd (approved):

Ostrzyciel (approved):

PR #271, opened 2026-08-06
Add missing slot_usage to anonymous_class_expression

### LINKED ISSUE #3851: Metamodel: AnonymousClassExpression inheritance references Definition instead of ClassDefinition
**Describe the bug**
[Class: AnonymousClassExpression](https://linkml.io/linkml-model/latest/docs/AnonymousClassExpression/) has inheritance-related `is_a`, but does not use the `slot_usage` that `ClassDefinition` uses, meaning that the inheritance slots simply refer to the broad `Definition` class, which seems unintended.

niegrzybkowski:
Closes https://github.com/linkml/linkml/issues/3851

### DIFF
--- a/linkml_model/model/schema/meta.yaml
+++ b/linkml_model/model/schema/meta.yaml
@@ -3084,6 +3084,10 @@ classes:
       - class_expression
     slots:
       - is_a
+    slot_usage:
+      is_a:
+        range: class_definition
+        description: A primary parent class from which inheritable metaslots are propagated
 
   class_definition:
     rank: 2


Ostrzyciel (approved):
looks good

linkml_model/model/schema/meta.yaml
```diff
@@ -3084,6 +3084,10 @@ classes:
       - class_expression
     slots:
       - is_a
+    slot_usage:
+      is_a:
+        range: class_definition
+        description: A primary parent class from which inheritable metaslots are propagated
```
matentzn:
```suggestion
        description: A class that any instance satisfying this expression must also be an instance of
```

See jsonschemagen:

> `is_a: <C>` is used in the context of an AnonymousClassExpression to include a constraint that instances of the expression must be instances of class `<C>`.

matentzn (commented):
Theoretically this is a breaking change - narrowing the definition of the slot. In practice, unless you have been using your own generators/validators, no one could have produced a schema  that will not pass the new validation. 

BUT: I want @gouttegd and @kervel on this too to agree here because you have both been building your own generators. Give me your god honest opinion.

Looks like this change is just confirming what owlgen, jsonschemagen and referencevalidator were assuming anyways. Cycling this through generators I get this:

- **OWL**: `allValuesFrom linkml:Definition` → `allValuesFrom linkml:ClassDefinition` on the `linkml:is_a` restriction of `AnonymousClassExpression`.
- **SHACL**: `sh:class linkml:Definition` → `sh:class linkml:ClassDefinition`.
- **Python**: `is_a: Optional[Union[str, DefinitionName]]` → `Optional[Union[str, ClassDefinitionName]]`, `__post_init__` coerces to `ClassDefinitionName`, plus a new additive `slots.anonymous_class_expression_is_a`. Both name types are `str` subclasses, so no runtime change.
- **JSON Schema**: *only the description changes*. `is_a` stays `"type": "string"` because it's a non-inlined name reference in both the old and new range.

gouttegd (dismissed):
(i) Breaking or not, as far as I can tell the change is correct.

(ii) This should have no effect whatsoever on the Java code generator in LinkML-Py.

(iii) I am not (yet) building my own code generator, just improving the one in LinkML-Py. :)

niegrzybkowski:
> Looks like this change is just confirming what owlgen, jsonschemagen and referencevalidator were assuming anyways.

That was the idea behind the PR: to make hidden generator assumptions described explicitly in the metamodel. This was an issue for us in LinkML-Scala, as the metamodel classes produced the wrong type of reference and made the slot unusable. I'll try to include more context in future PRs

cmungall (approved):
Good catch. As noted, this was implicit anyway.

kervel:
no impact from my side!

PR #274, opened 2026-08-27
Add `allow_unknown_designator` slot.

gouttegd:
This PR implements the idea proposed in https://github.com/linkml/linkml/issues/3701 (and refined after a meeting with the core devs as outlined [here](https://github.com/linkml/linkml/issues/3701#issuecomment-5428760611)).

That is, it adds a new slot (applicable to a slot definition) to alter the behaviour of a type designator slot.

If a slot is marked with both:

```yaml
designates_type: true
allow_unknown_designator: true
```

then the constraint that the value of the slot must match the name/URI/CURIE of a known subclass of the class defining the slot (standard constraint for a type designator slot) is relaxed.

While we are at it, the description of the `designates_type` slot is also updated to make its meaning hopefully clearer. We also warn against the use of **multi-valued** type designators, as per https://github.com/linkml/linkml/issues/1681.

### DIFF
--- a/linkml_model/model/schema/meta.yaml
+++ b/linkml_model/model/schema/meta.yaml
@@ -1932,7 +1932,7 @@ slots:
     range: boolean
     inherited: true
     description: >-
-      True makes the slot the “type designator slot” for its class. A type designator slot is used to determine the runtime type
+      True makes the slot the "type designator slot" for its class. A type designator slot is used to determine the runtime type
       of an object.
 
       When the type designator slot is present in an object, its value MUST match either the name, the URI, or the CURIE
@@ -1967,7 +1967,7 @@ slots:
         base class) while ensuring that data conformant to the extended schema can still be considered valid against the original
         schema.
       - >-
-        In most cases, schema authors will want to make sure that the base class is configured to allow “extra slots”, so that
+        In most cases, schema authors will want to make sure that the base class is configured to allow "extra slots", so that
         any slot defined by an unknown subclass can be accepted without triggering an error.
     in_subset:
       - SpecificationSubset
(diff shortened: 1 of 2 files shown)


linkml_model/model/schema/meta.yaml
```diff
@@ -1932,11 +1932,49 @@ slots:
...
+      - >-
+        Setting this slot to True only makes sense if the base class is NOT abstract; if the base class is abstract, then by
+        definition it cannot be instantiated, so an object cannot be assumed to be an instance of the base class.
+      - >-
+        This slot is intended to allow a schema to be extended by another schema (that could add new subclasses to the original
+        base class) while ensuring that data conformant to the extended schema can still be considered valid against the original
+        schema.
+      - >-
+        In most cases, schema authors will want to make sure that the base class is configured to allow "extra slots", so that
+        any slot defined by an unknown subclass can be accepted without triggering an error.
```
gouttegd, line 1971: Alternatively, we could decide that the mere presence of `allow_unknown_designator` is enough to imply that the class carrying the type designator slot is allowing extra slots (this is a possibility that was discussed in the aforementioned meeting).

If we wanted to do that, then (i) the last comment above should be removed, and (ii) the description should be amended to add:

> Additionally, when this slot is set to `True`, the class carrying the type designator slot should be implicitly considered to allow extra slots (as if it was defined with `extra_slots.allowed=True`), unless the class definition has an explicit `extra_slots` configuration.

(For what it’s worth, I am not convinced this would be a good idea, if only because it creates “long range effects”: a class could be implicitly allowing extra slots even if nothing in the class definition says so, just because it has a slot – defined elsewhere in the schema – that has `allow_unknown_designator: True`. Better in my opinion to be explicit.)

niegrzybkowski: My unprompted 2 cents: skip the "allow_unknown_designator => extra_slots" implication. The subclasses could have the same slots as the base class, but may define different behaviors in their slot_usages, like expressions or rules.

PR #279, opened 2026-09-16
Release prep: generate v1.12.0-rc1 against linkml main

amc-corey-cox:
Release prep for **v1.12.0-rc1** so linkml can vendor and test against it. Does not cut the release.

The committed artifacts are stale against `meta.yaml` (#270/#271 changed it without a regeneration), and they don't reproduce under the next linkml either — regenerating with `main` rather than 1.11.1 changes `rdf/*.ttl` (linkml#3641), a `meta.py` type, and adds a `java/` tree. So this generates against `main`.

- `tool.uv.sources` pins linkml + linkml-runtime to `063d4665f`. Dev-side only; the wheel still carries `project.dependencies`.
- Exact pins became floors — a `main` checkout reports `1.11.0.post365.dev0+063d4665`, which can't satisfy `==1.11.1`.
- `java` excluded from gen-project: javagen joined its defaults, and nothing can address the output (`Format` has no `JAVA` member) — 44 files that would also land in the linkml-runtime wheel via `make update_model`.
- `metamodel_version` → 1.12.0, plus full regeneration.

**Two notes for review:**

`metamodel_version` lands unevenly — `meta.py` gets 1.12.0 from `meta.yaml`, the other modules get 1.11.0 from linkml-runtime's vendored metamodel (still model 1.11.1-rc1). Resolves once linkml vendors this rc.

The ~26k-line diff is almost entirely timestamps and RDF/SQL ordering. Checked by graph comparison: SHACL is semantically identical bar the `is_a` change (5,135 property shapes, zero differences); sorted SQL leaves only the real changes. The content is #271 propagating — `AnonymousClassExpression.is_a` narrowing to `ClassDefinition` across every artifact.

**Before final release:** swap the pin and floors for the released versions, regenerate, confirm no change.

Verified: `make gen-project`, `make test` (8 passed, 5 skipped), `pre-commit run --all-files` clean.


linkml_model/jsonld/meta.jsonld
```diff
@@ -6771,11 +6797,10 @@
...
       "definition_uri": "https://w3id.org/linkml/Anything",
       "from_schema": "https://w3id.org/linkml/meta",
-      "mappings": [
+      "exact_mappings": [
         "linkml:Any"
       ],
       "slot_usage": {},
-      "class_uri": "https://w3id.org/linkml/Any",
       "@type": "ClassDefinition"
     },
```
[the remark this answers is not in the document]
amc-corey-cox: Correct observation, but this is intentional rather than lossy. linkml#2687 argued that serialising `class_uri` as `skos:mappingRelation` was too fuzzy and should be `skos:exactMatch` — which is what the OWL output already emitted. That was fixed in linkml `2b8a0241e` ("Fix 2687: Make `class_uri` be `skos:exactMatch` of the element URI on JSON-LD and RDF"), so this change aligns JSON-LD/RDF with OWL rather than diverging from it.

It surfaces in this PR because the commit is not an ancestor of `v1.11.1` — it landed on main after that release was cut, and this PR pins generation to linkml main. The Python API is unaffected (39 `class_class_uri` entries before and after); the change is confined to the JSON-LD/RDF serializations.

The downstream point is noted separately: consumers reading `class_uri` as a key out of `meta.jsonld` will no longer find it, so that belongs in the v1.12.0 release notes.

pyproject.toml
```diff
@@ -27,15 +27,23 @@ classifiers = [
...
 Repository = "https://github.com/linkml/linkml-model/"
 Documentation = "https://linkml.io/linkml-model/docs/"
 
+[tool.uv.sources]
+# Release-prep pin: generate this rc's artifacts with the linkml commit that is
+# expected to become the next linkml release. Dev-side only -- uv strips sources
+# from built metadata, so the published wheel carries project.dependencies as-is.
+# Replace with the released versions before cutting the final release.
+linkml = { git = "https://github.com/linkml/linkml", rev = "063d4665f", subdirectory = "packages/linkml" }
+linkml-runtime = { git = "https://github.com/linkml/linkml", rev = "063d4665f", subdirectory = "packages/linkml_runtime" }
```
kevinschaper: Pinning to `rev` rather than a branch is the right call — this stays reproducible, and the comment explains itself well.

The part that worries me is the paired change in `[project] dependencies`: `linkml==1.11.1` → `linkml>=1.11.0,<2`. Unlike the source pin, that one *does* ship in the wheel, so anyone installing the rc can resolve any linkml ≥1.11.0 — including one whose vendored metamodel predates this model. The reasoning for it is sound (a `main` checkout reports `1.11.0.post365.dev0+063d4665`, which can't satisfy `==1.11.1`), but "swap the pin and floors back before final release" is now a manual step in a PR description. Worth an issue or a release-checklist entry so it can't be missed — forgetting it ships the loose floor permanently.

linkml_model/meta.py
```diff
@@ -81,7 +80,7 @@
 from .units import UnitOfMeasure
 from linkml_runtime.utils.metamodelcore import Bool, NCName, URI, URIorCURIE, XSDDateTime
 
-metamodel_version = "1.11.0"
+metamodel_version = "1.12.0"
```
kevinschaper: Confirmed the unevenness you flagged — `meta.py` is the only module that moves:

```
meta.py          1.12.0
annotations.py   1.11.0
mappings.py      1.11.0
types.py         1.11.0
units.py         1.11.0
validation.py    1.11.0
extensions.py    1.11.0
datasets.py      1.11.0
```

Agreed it resolves once linkml vendors this rc, since the other seven read the version from linkml-runtime's vendored copy. Flagging only so it's an explicit confirm-item on the rc1 → rc2 pass rather than something noticed later: if it hasn't evened out after vendoring, the vendor step didn't take.

pyproject.toml
kevinschaper: Pinning to a `rev` rather than a branch is the right call, and the comment explains itself well.

The part worth a safety net is the paired change in `[project] dependencies`: `linkml==1.11.1` → `linkml>=1.11.0,<2`. Unlike the source pin, that one ships in the wheel, so an rc installer can resolve any linkml ≥1.11.0 — including one whose vendored metamodel predates this model. The reasoning is sound (a `main` checkout reports `1.11.0.post365.dev0+063d4665`, which can't satisfy `==1.11.1`), but "swap the pin and floors back before the final release" currently lives only in the PR description. Worth an issue or a release-checklist entry — forgetting it ships the loose floor permanently.

linkml_model/meta.py
kevinschaper: Confirmed the unevenness you flagged — `meta.py` is the only module that moves:

```
meta.py          1.12.0
annotations.py   1.11.0
mappings.py      1.11.0
types.py         1.11.0
units.py         1.11.0
validation.py    1.11.0
extensions.py    1.11.0
datasets.py      1.11.0
```

Agreed it resolves once linkml vendors this rc, since the other seven take the version from linkml-runtime's vendored copy. Flagging it only so it becomes an explicit confirm-item on the rc1 → rc2 pass: if it hasn't evened out after vendoring, the vendor step didn't take.

kevinschaper (approved):
no surprises found, looks good!

linkml_model/jsonschema/meta.schema.json
```diff
@@ -12,7 +12,7 @@
             "type": "string"
         },
         "AltDescription": {
-            "additionalProperties": false,
+            "additionalProperties": true,
```
matentzn, line 15: This one isn't cosmetic. Is it intended? 58 classes in this file flip `"additionalProperties": false` -> `true`.

It comes from the new `extra_slots` implementation in jsonschemagen. `get_additional_properties()` returns `self.not_closed` when a class has no `extra_slots`, and `not_closed` defaults to `True`. The old code hardcoded `additional_properties = False` for any class that wasn't unconstrained. So the per-class default inverted.

To me it seems this contradicts our own metamodel. `meta.yaml` documents `extra_slots` as "`allowed: false` (or `allowed:`/`allowed: null` while `range_expression` is `null`) - forbid all additional data **(default)**". The generator does the opposite when the slot is absent. Furthermore, our artifacts now disagree with each other. I canonicalised `meta.shacl.ttl` before and after and every node shape keeps its `sh:closed true`. So as of this rc, validating a schema against the SHACL rejects unknown keys and validating the same schema against this JSON Schema accepts them. Anyone using `meta.schema.json` to lint schemas silently stops catching `descrption:` / `slot_usge:` typos.

linkml_model/meta.py
```diff
@@ -2996,7 +2995,7 @@ class ArrayExpression(YAMLRoot):
 
     exact_number_dimensions: Optional[int] = None
     minimum_number_dimensions: Optional[int] = None
-    maximum_number_dimensions: Optional[Union[dict, Anything]] = None
+    maximum_number_dimensions: Optional[Union[int, bool, Bool]] = None
```
matentzn, line 2998: Nit, but flagging it because it's an API change rather than a timestamp: this narrows from `Optional[Union[dict, Anything]]` to `Optional[Union[int, bool, Bool]]` and adds coercion, so a value that previously passed through untouched can now raise `ValueError`.

I'm not worried in practice - the slot is `status: testing`, and a real `False` stays `False` since `isinstance(False, int)` short-circuits the coercion. Just shouldn't go out unmentioned.

matentzn (commented):
I have some nits as well, but for now, my two most pressing concerns.

linkml_model/jsonschema/meta.schema.json
amc-corey-cox: Confirmed — reproduced all of it, including the counts: `false` 58 → 0.

**Where it changed:** linkml/linkml#3612 (`aeef7748b`, 18 Jun) turned the `extra_slots`-absent branch from `return False` into `return self.not_closed`. It was fixing a real bug — linkml/linkml#3611, where `not_closed` was ignored for `$defs` — but since `not_closed` defaults to `True`, making it authoritative everywhere also inverted the per-class default.

**Where we'd said otherwise:** linkml-model#205, which introduced `extra_slots` — *"By default linkml forbids all additional data."* `meta.yaml:1376` still documents that, and `shaclgen` (`closed: bool = True`) and `javagen` both still follow it, which is why your SHACL point holds: 34 `sh:closed true`, unchanged.

**Fix** (linkml/linkml#4021): `not_closed` goes back to a plain bool defaulting to `False`, governing classes only. The top level comes from the root class instead — `--top-class` or `tree_root: true` — and stays open when there is no root class, since a property-less closed top level would admit nothing but `{}`.

I've re-pinned this PR to the branch carrying it and regenerated: `meta.schema.json` is back to 58 closed and 2 open, matching what ships today, and it catches a misspelled `descrption:` again. Biolink's 337 classes likewise go back to closed.

One change beyond restoring the old behaviour: a schema with a `tree_root` now closes its top level too, so an unknown top-level key is rejected where it previously passed.

Pinning to an unmerged branch is deliberate for an rc — what matters is that the two agree at release. Leaving this open until linkml/linkml#4021 merges and the pin moves to main.

linkml_model/meta.py
amc-corey-cox: Agreed on all of it, including the reasoning — `isinstance(False, int)` short-circuits before the coercion at `meta.py:3043`, so a genuine `False` is never touched. What changed is that the slot now carries a coercion block that raises `ValueError` when neither `int()` nor `Bool()` succeeds, where previously any value passed through untouched.

No change here — it follows from `array.yaml`, not from this PR's regeneration. But you're right that it shouldn't go out unmentioned, so it's going in the v1.12.0 release notes along with the other API-visible change in this regeneration: `class_uri` now serialises as `skos:exactMatch` in JSON-LD and RDF rather than as a `class_uri` key (linkml/linkml#2687), so anything reading `class_uri` out of `meta.jsonld` will no longer find it.

amc-corey-cox:
@kevinschaper your approval got auto-dismissed by a push, sorry — nothing you did.

One commit since: re-pinned to linkml/linkml#4021 and regenerated `meta.schema.json`, which fixes the open-by-default classes Nico caught. Needs two approvals here, so a re-review when you get a chance.

